PR #121, opened 2025-08-11
fix: avoid task not found bug

luoyou:
### Description

When user send task and click the pause button immediately, it will trick Task not found error.
move create_task_lock to top postion.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


a7m-1st:
Thanks for the PR @luoyou. Currently when toggling the task it gets locked like below. Does it target this problem?

image

Wendong-Fan (approved):

PR #298, opened 2025-09-03
feat: i18n

ClearLuvMoki:
### Description

 Add some locales

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


a7m-1st:
Hi there, thanks @ClearLuvMoki for your PR. I support the need for internationalization, but I have some comments

src/i18n/locales/en-US.json
a7m-1st, line 1: To begin with, if not mistaken all translation json is recommended to be under /public in the format of:
```
public/               # if using create-react-app or Vite
 │  └─ locales/
 │       ├─ en/
 │       │   └─ translation.json
 │       ├─ fr/
 │       │   └─ translation.json
 │       └─ ...
 ```

File: src/i18n/index.ts
Comment (a7m-1st) [inline src/i18n/index.ts:1]: I recommend the config file to be under `./src/configs/i18next.config`, so it would be cleaner.

File: package.json [hunk 1]
```diff
@@ -95,10 +95,12 @@
     "autoprefixer": "^10.4.20",
     "electron": "^33.2.0",
     "electron-builder": "^24.13.3",
+    "i18next": "^25.4.2",
     "postcss": "^8.4.49",
     "postcss-import": "^16.1.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
+    "react-i18next": "^15.7.3",
     "tailwindcss": "^3.4.15",
```
a7m-1st, line 104: electron-menu gif
  Source (deprecated): https://phrase.com/blog/posts/building-an-electron-app-with-internationalization-i18n/

Lastly thanks for choosing those libraries, but I couldn't recommend enough `i18next-fs-backend`, where `react-i18next` actually doesn't have access to the file system as nodejs/electron has. Maybe react-i18next for React based components, and fs-backend for electron components.

Then if not mistaken we can also use electron event handlers with it:

```js
const i18n = require('./src/configs/i18next.config');
const menuFactoryService = require('./src/services/menuFactory');
...
function createAppWindow() {
...
i18n.on('loaded', (loaded) => {
    i18n.changeLanguage('en');
    i18n.off('loaded');
  });
i18n.on('languageChanged', (lng) => {
    menuFactoryService.buildMenu(app, win, i18n);
  });
  ```
  
  I believe we just need to test macos, bcz I am worried that if we are showing the menus in the system top bar, then react can't do anything about it.

a7m-1st (changes requested):
Do let me know if you have any comments that I overlooked or if you need help. Thanks.

package.json
ClearLuvMoki: If there is too much internationalized text in the main process of electron, can choose to use i18next-fs-backend，share the locale file with the front end

a7m-1st: All right cool, lets keep note of that perhaps once the react implementation is solid.

a7m-1st:
Lastly @ClearLuvMoki , I forgot to mention, by having empty cn translations.json but the config already present I think when the system is in Chinese then the `t(headerText)` will just appear as **headerText**.

So its better to either remove the config, or populate the other translations than leave it empty.

ClearLuvMoki:
> Lastly @ClearLuvMoki , I forgot to mention, by having empty cn translations.json but the config already present I think when the system is in Chinese then the `t(headerText)` will just appear as **headerText**.
> 
> So its better to either remove the config, or populate the other translations than leave it empty.

@a7m-1st Thanks for the reminder. First of all, this part is not finished yet. Secondly, there is no method to call to change the language, so the language will not be Chinese.

a7m-1st:
> > Lastly @ClearLuvMoki , I forgot to mention, by having empty cn translations.json but the config already present I think when the system is in Chinese then the `t(headerText)` will just appear as **headerText**.
> > So its better to either remove the config, or populate the other translations than leave it empty.
> 
> @a7m-1st Thanks for the reminder. First of all, this part is not finished yet. Secondly, there is no method to call to change the language, so the language will not be Chinese.

All right cool, thanks for your reply @ClearLuvMoki

FooFindBar:
Hi @ClearLuvMoki ,

Thank you for your PR. I have added a language switch method on the [pr-feat-i18n](https://github.com/eigent-ai/eigent/tree/pr-feat-i18n) branch, based on your PR. You can take a look for reference or merge it directly if you like.

PR #332, opened 2025-09-09
ix: resolve issue where tasks from previous account kept running afte…

FooFindBar:
…r switching accounts

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


electron/main/index.ts
```diff
@@ -312,6 +312,13 @@ function registerIpcHandlers() {
   });
   ipcMain.handle('get-app-version', () => app.getVersion());
   ipcMain.handle('get-backend-port', () => backendPort);
+  ipcMain.handle('restart-backend', () => {
+    if (backendPort) {
+      cleanupPythonProcess();
+      checkAndStartBackend();
+    }
+    return { success: true };
+  });
```
a7m-1st: Thanks for the PR @FooFindBar , but cleanup & check functions are async. Wouldn't this cause race conditions if targeting the same port?

I suggest to:
```ts
 ipcMain.handle('restart-backend', async () => {
  try {
    if (backendPort) {
      await cleanupPythonProcess();
      await checkAndStartBackend();
    }
    return { success: true };
    } catch(error) {
      log.error('Error occurred while restarting the backend:', error);
      return { success: false, error: error };
    }
  });
```

src/store/chatStore.ts
```diff
@@ -1639,6 +1639,7 @@ const chatStore = create<ChatStore>()(
 			const { create } = get()
 			console.log('clearTasks')
 			fetchDelete('/task/stop-all')
+			window.ipcRenderer.invoke('restart-backend')
```
a7m-1st, line 1642: Then we can handle the promise down here if required.

https://www.electronjs.org/docs/latest/api/ipc-renderer#ipcrendererinvokechannel-args

a7m-1st (commented):
Just concerned about the async nature of cleanupPythonProcess() and checkAndStartBackend()

a7m-1st (commented):
Thanks for handling the exception @Wendong-Fan

PR #337, opened 2025-09-12
chore: Slack enhance support file_path

fengju0213:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Wendong-Fan (commented):
thanks @fengju0213 , I think this change requires camel's version been updated, otherwise would cause error

Issue #237, opened 2025-08-27
[Feature Request] Integrate Whatsapp


Pakchoioioi:
### Motivation
test and integrate camel's WhatsAppToolkit into Eigent, user could config from the Tools section as shown in the image below

<img width="669" height="182" alt="Image" src="https://github.com/user-attachments/assets/432dd8e0-fdaf-4fc0-93ff-e6def357b208" />

Additionally, we need to support the ability for users to trigger eigent tasks by sending messages directly through WhatsApp.

fury0928:
Hi @Wendong-Fan, hope you’re having a good weekend. 
Could you please add a bit more detail here so I can pick this up?

Wendong-Fan:
> Hi [@Wendong-Fan](https://github.com/Wendong-Fan), hope you’re having a good weekend. Could you please add a bit more detail here so I can pick this up?

Hi @eureka928,

We already have a basic WhatsApp toolkit wrapping CAMEL's [WhatsAppToolkit](https://github.com/camel-ai/camel/blob/master/camel/toolkits/whatsapp_toolkit.py) — it can send messages, get templates, and fetch business profiles. Users configure it with WHATSAPP_ACCESS_TOKEN and WHATSAPP_PHONE_NUMBER_ID.

The missing part is inbound messaging — receiving WhatsApp messages to trigger Eigent tasks. This should be built as a new trigger type (TriggerType.WhatsApp) on top of the trigger system being introduced in #823.

So I'd suggest waiting for #823 to land first, then:

Register a webhook endpoint with Meta to receive incoming WhatsApp messages (including the GET verification handshake)
Add a WhatsApp trigger type with its config schema (access token, phone number ID, verify token, etc.)
Parse incoming messages → create trigger executions
Return task results to the user via the existing send_message
Happy to discuss further once #823 is merged.

a7m-1st:
After a quick read regarding this architecture, the way I usually do is create a schedule trigger, then use the toolkit via an agent to get for messages. But this comes with an overhead; so a better solution would be to develop some kind of middleware that uses the get_messages tool programatically.

I think it would be clear once I make a PR regarding the server update. We have all the pieces required to make the middleware now:
- Triggers, Task Queues, Execution Listeners

dataCenter430:
Hi, @Pakchoioioi  @Wendong-Fan  I'm confident that I can integrate Whatsapp into Eigent perfectly.
Could you please assign it to me?

[...]

PR #339, opened 2025-09-12
enhance: mcp config page

a7m-1st:
### Description
Fixing #237 was a rabbit hole, still couldn't pinpoint the issue. Along the way here are some frontend fixes I made:
- MCP user api server config & mcp.json was not in sync
- Fixed duplicate server names
- disabled edit names because the mcp.json would think its a new server (i.e. updated based on name)

Points along the way:
- The error got nothing to do with electron, its actually not configuring the backend right
- The route `@router.post("/chat/{id}/install-mcp")` seems to be ignored and configured directly in mcp_agent creation. I.e. the function where 'mcp.get_tools()' isn't getting added to mcp_agent (/agent.py).
- If you run 'Whatsapp` mcp with `/examples/mcp/mcp_toolkit.py` it works fine ! ✅ I couldn't see any error logs in the backend, frontend or the /logs 
- While editing mcp config files, make sure to double check mcp.json as the parsing may be skewed for args

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/component/environment.py
+++ b/backend/app/component/environment.py
@@ -5,10 +5,36 @@ from fastapi import APIRouter, FastAPI
 from dotenv import load_dotenv
 import importlib
 from typing import Any, overload
+import threading
+
+# Thread-local storage for user-specific environment
+_thread_local = threading.local()
+
+# Default global environment path
+default_env_path = os.path.join(os.path.expanduser("~"), ".eigent", ".env")
+load_dotenv(dotenv_path=default_env_path)
 
 
-env_path = os.path.join(os.path.expanduser("~"), ".eigent", ".env")
-load_dotenv(dotenv_path=env_path)
+def set_user_env_path(env_path: str | None = None):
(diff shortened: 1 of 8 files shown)


a7m-1st:
Fun fact, I also tried to set a streaming http based WA mcp, hosted locally with fastmcp. McpClient supports http, sse or websocket connections but the api in https://dev.eigent.ai/api/mcp/import/local expects `command & args` fields only thus fails. 

```json
"whatsapp-server": {
  "url": "http://localhost:8087/mcp"
}
```

FooFindBar (approved):

a7m-1st:
Thanks for the Review @FooFindBar, I have fixed args parsing which should make editing almost end to end without needing to touch `mcp.json`.

Changes:
Fix: 
image

Handle args:
image

a7m-1st:
closes #346

Issue #327, opened 2025-09-08
[BUG] the task from the right is in progress, but the task on the left is done


Pakchoioioi:
### What version of eigent are you using?

0.0.65

### System information

mac m chip

### Problem description

the task from the right is in progress, but the task on the left is done

### log

<img width="2224" height="1706" alt="Image" src="https://github.com/user-attachments/assets/e72cc907-4bb6-437b-b886-8d1396125428" />

Issue #322, opened 2025-09-07
[BUG] Task flow exception and incorrect progress display


Pakchoioioi:
### What version of eigent are you using?

0.0.65

### System information

mac m chip

### Problem description

The Notion agent fails and the task is assigned to the developer agent but the task is marked as done The developer did not finish the task but there is 4/4 done in the panel

<img width="1436" height="1846" alt="Image" src="https://github.com/user-attachments/assets/f140dac9-a840-4826-8a7e-744435be1ad1" />
<img width="1436" height="1846" alt="Image" src="https://github.com/user-attachments/assets/3a94eb0b-a3e7-4db0-a0af-a9869696fe88" />

### log

[eigent-0.0.65-darwin-arm64-1757203468028.log](https://github.com/user-attachments/files/22194636/eigent-0.0.65-darwin-arm64-1757203468028.log)

Issue #110, opened 2025-08-08
[Feature Request] When a task assigned to a worker is not completed and gets reassigned to another agent, the task log needs to be optimized - need to be confirmed


Pakchoioioi:
### What version of eigent are you using?

0.0.44

### System information

all os

### Problem description

When a task assigned to a worker is not completed and gets reassigned to another agent, the task log needs to be optimized. Currently it just print as the same which is not make sense.

**Expected behavior:**

Worker 1 executes the task: subtasks + log contents

Worker 1 fails to complete the task and it is reassigned: subtasks (Attempt 1) + log contents + task failed

Reassigned to Worker 2 for execution:
Worker 2 receives the task: subtasks (Attempt 2) + log contents

<img width="1670" height="1066" alt="Image" src="https://github.com/user-attachments/assets/0d5139a5-718d-4489-8363-188f4ab5df4f" />

### log

[eigent-0.0.44-darwin-arm64-1754650287602.log](https://github.com/user-attachments/files/21681999/eigent-0.0.44-darwin-arm64-1754650287602.log)

Pakchoioioi:
Will be modified at same time: https://github.com/eigent-ai/eigent/issues/322

Douglasymlai:
Design File for new subtask status classification and tags.

<img width="706" height="409" alt="Image" src="https://github.com/user-attachments/assets/d270b22f-a8ed-4ad3-ae78-89d5fb827591" />

https://www.figma.com/design/X61oO7bMCb7u9GVrk2cCfR/Eigent-Local-Device?node-id=5382-53484&t=DUY0VG2eX8ZpgpoC-1

PR #341, opened 2025-09-15
Fix/task log reassign incomplete #110 #327

FooFindBar:

- [x] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

#110
#327
#322


src/components/WorkFlow/node.tsx
```diff
@@ -543,53 +601,90 @@ export function Node({ id, data }: NodeProps) {
...
-											)}
-											{task.status === "blocked" && (
-												<TriangleAlert
-													size={16}
-													className="text-icon-warning"
-												/>
-											)}
-											{(task.status === "" || task.status === "waiting") && (
-												<Circle size={16} className="text-slate-400" />
+										{task.status}
```
Wendong-Fan: seems this need to be cleaned?

PR #360, opened 2025-09-16
fix: mcp agent not calling expected tools & add logs

a7m-1st:
### Description

- Whatsapp mcp wasn't working due to Mcp agent being commented out
- Mcp_agent was not using whatsapp mcp due to using `search` tool and a hidden error of "env config empty"
image

- Added mcp client termination handling & logs

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


backend/app/service/chat_service.py
```diff
@@ -437,11 +437,11 @@ async def construct_workforce(options: Chat) -> tuple[Workforce, ListenChatAgent
...
-    # workforce.add_single_agent_worker(
-    #     "MCP Agent: A Model Context Protocol agent that provides access "
-    #     "to external tools and services through MCP integrations.",
-    #     mcp,
-    # )
+    workforce.add_single_agent_worker(
+        "MCP Agent: A Model Context Protocol agent that provides access "
+        "to external tools and services through MCP integrations.",
+        mcp,
+    )
```
Wendong-Fan: thanks @a7m-1st , I think our system now support config the mcp via adding a new worker agent, so there's no need to add this special MCP Agent, could you double check this?

a7m-1st: oh okay, thanks for the review @Wendong-Fan, let me check

a7m-1st: image

- @Wendong-Fan , I have just reverted that code & adding a separate worker is working fine. I will make a another PR for some enhancements on Add worker along the way.
- So I guess this pr is only about enhanced mcp logging & mcp client termination handling.

backend/app/utils/agent.py
```diff
@@ -1475,12 +1480,29 @@ async def get_mcp_tools(mcp_server: McpServers):
...
         traceroot_logger.error(f"Failed to connect MCP toolkit: {e}", exc_info=True)
-    return mcp_toolkit.get_tools()
+        return []
+    finally:
+        # Ensure proper cleanup without causing scope issues
+        if mcp_toolkit is not None:
+            try:
+                # Don't explicitly close here as it may cause scope issues
+                # Let the toolkit handle its own cleanup
+                pass
```
Wendong-Fan: seems this finally block would do nothing?

a7m-1st: The toolkit does the cleanup internally

PR #407, opened 2025-09-28
enhance: Fix #352 install enhanced PR404

### LINKED ISSUE #352: [Feature Request] Improve the robustness of installation
### Motivation

Now the desktop app installation is not so stable
1. sometimes the install finished in the backend, but the frontend is still loading
2. sometimes after long time hanging in the frontend, user clicked "retry", the desktop app loaded into the main interface, but the install may still failed, which cased the app couldn't run
3. We didn't show the error message for the installation in the frontend, which makes it confusing for the user

we need to test and validate the install logic comprehensively to make sure the install is reliable

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

Wendong-Fan:
### Description

  1. Removed unused import - Deleted import e from 'express' in electron/main/index.ts:21
  2. Fixed type inconsistency - Updated code to handle checkToolInstalled() returning PromiseReturnType instead of boolean
  3. Improved error handling - Converted installCommandTool() from Promise constructor to async/await with proper try-catch
  4. Fixed UV path configuration - Updated uv() function path from ~/.eigent/bin/uv to ~/.local/bin/uv in
  backend/app/component/command.py:9

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/install-deps.ts
+++ b/electron/main/install-deps.ts
@@ -100,7 +100,7 @@ Promise<PromiseReturnType> => {
  * Check if command line tools are installed, install if not
  */
 export async function installCommandTool(): Promise<PromiseReturnType> {
-  return new Promise(async (resolve, reject) => {
+  try {
       const ensureInstalled = async (toolName: 'uv' | 'bun', scriptName: string): Promise<PromiseReturnType> => {
         if (await isBinaryExists(toolName)) {
             return { message: `${toolName} already installed`, success: true };
@@ -123,24 +123,26 @@ export async function installCommandTool(): Promise<PromiseReturnType> {
           });
         }
 
-        return { 
+        return {
           message: installed ? `${toolName} installed successfully` : `${toolName} installation failed`,
-          success: installed 
+          success: installed
         };
       };
 
       const uvResult = await ensureInstalled('uv', 'install-uv.js');
       if (!uvResult.success) {
-          return reject({ message: uvResult.message, success: false });
-      }
-      
-      const bunResult = await ensureInstalled('bun', 'install-bun.js');
-      if (!bunResult.success) {
-          return reject({ message: bunResult.message, success: false });
+          return { message: uvResult.message, success: false };
       }
 
-      return resolve({ message: "Command tools installed successfully", success: true });
-  })
+      const bunResult = await ensureInstalled('bun', 'install-bun.js');
+      if (!bunResult.success) {
+          return { message: bunResult.message, success: false };
+      }
+
+      return { message: "Command tools installed successfully", success: true };
+  } catch (error) {
+      return { message: `Command tool installation failed: ${error}`, success: false };
+  }
 }
 
 let uv_path:string;
--- a/backend/app/component/command.py
+++ b/backend/app/component/command.py
@@ -6,4 +6,4 @@ def bun():
 
 
 def uv():
-    return os.path.expanduser("~/.local/bin/uv")
+    return os.path.expanduser("~/.eigent/bin/uv")
--- a/backend/app/component/command.py
+++ b/backend/app/component/command.py
@@ -6,4 +6,4 @@ def bun():
 
 
 def uv():
-    return os.path.expanduser("~/.eigent/bin/uv")
+    return os.path.expanduser("~/.local/bin/uv")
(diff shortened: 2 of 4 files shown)


electron/main/install-deps.ts
```diff
@@ -100,7 +100,7 @@ Promise<PromiseReturnType> => {
  * Check if command line tools are installed, install if not
  */
 export async function installCommandTool(): Promise<PromiseReturnType> {
-  return new Promise(async (resolve, reject) => {
+  try {
       const ensureInstalled = async (toolName: 'uv' | 'bun', scriptName: string): Promise<PromiseReturnType> => {
```
a7m-1st, line 104: Great Catch.  Looked into it & you are right, `new Promise` is unnecessary here & in many others. Its just a js way to create manual low-level async for non awaitable functions. 

These days TypeScript automatically wraps the return value in a Promise for async funcs.

```ts
// Even though you return a string, the function returns Promise<string>
async function getName(): Promise<string> {
  return "John"; // TypeScript wraps this in Promise.resolve("John")
}

// This is equivalent to:
function getNameManual(): Promise<string> {
  return Promise.resolve("John"); //& ya, static function, no need to create class
}
```

Thanks @Wendong-Fan , LGTM.

FooFindBar:
@a7m-1st I’m unclear about the logic in two places:
https://github.com/eigent-ai/eigent/blob/453800032e8e0b2b6e8bf40f1fd66d5a84aac8c6/src/hooks/useInstallationSetup.ts#L19C3-L47C52
~~~ ts
useEffect(() => {
    const checkToolInstalled = async () => {
      try {
        console.log('[useInstallationSetup] Checking tool installation status...');
        const result = await window.ipcRenderer.invoke("check-tool-installed");
        
        if (result.success && initState === "done" && !result.isInstalled) {
          console.log('[useInstallationSetup] Tool not installed, setting initState to carousel');
          setInitState("carousel");
        }        
      } catch (error) {
        console.error("[useInstallationSetup] Tool installation check failed:", error);
      }
    };

    const checkBackendStatus = async() => {
      // Also check if installation is currently in progress
      const installationStatus = await window.electronAPI.getInstallationStatus();
      console.log('[useInstallationSetup] Installation status check:', installationStatus);
      
      if (installationStatus.success && installationStatus.isInstalling) {
        console.log('[useInstallationSetup] Installation in progress, starting frontend state');
        startInstallation();
      }
    }
    
    checkToolInstalled();
    checkBackendStatus();
  }, [initState, setInitState, startInstallation]);
~~~
1. The **checkToolInstalled** method only calls **setInitState("carousel")**, but it doesn’t start installing dependencies or show the dependency installation page. It seems that dependency installation is controlled via startInstallation, not setInitState.

2. The **checkBackendStatus** method calls **startInstallation** when the tool is not installed. However, after the tool is installed, the backend service does not seem to start.

a7m-1st:
1. As for `checkToolInstalled()`, yes you are right. It just sets the installation page to type of "carousel" rather than without it. I didn't change this code as it was there in `Layout/index.tsx` previously. Thus when installing `without` the carousel it shows:
Screenshot 2025-08-12 144606

2. To be more exact, I didn't add any update logic to the backend. Its only a Ping status to see if the `Installing.lock` is free or no in the backend (Read only). That is because relying on the event listners to notify `install-dependencies-start` as the only source of truth is not reliable due to race conditions.

& Honestly this is the previous flow:
image

### But in reality here is what is happening:
- All the old logic in backend logic is the same, except that I removed the `frontend-ready` event which was trying to fix this frontend issue. 
- The only significant change I made is that now, the frontend doesn't wait for dependencies-start to show the Installation page, it can detect it anytime if the lock file is Installing & display the page.
- Thus this has allowed for checking for `uv sync` installations on runtime by `detectInstallationLogs(msg:string)`
image

a7m-1st:
@FooFindBar  So to recap about 2:
- I think its just a naming issue, all it does is show the frontend. I assume the backend always starts installation correctly.
```ts
//InstallationStore.ts
// Basic actions
startInstallation: () => 
  set({
    state: 'installing',
    progress: 20,
    logs: [],
    error: undefined,
    isVisible: true,
  }),
```
- Most backend logic is the same. I just refactored it for easier maintenance & error response. So any updates to enhance it is welcome, as I couldn't find a reliable way / library to handle this for us, except for using logs.

Thanks for your questions !

PR #426, opened 2025-10-01
enhance: isloate install environment

Wendong-Fan:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


electron/main/index.ts
```diff
@@ -919,14 +936,153 @@ async function createWindow() {
...
+      }
+    } catch (error) {
+      log.error('Error clearing localStorage:', error);
+    }
+
+    // Set up the injection for when page loads
+    win.webContents.on('dom-ready', () => {
+      log.info('DOM ready - creating auth-storage with carousel state');
+      win!.webContents.executeJavaScript(`
+        (function() {
```
a7m-1st, line 1003: Perhaps we can check for win before running event listener?
```js
if (!win || win.isDestroyed()) {
        log.warn('Window destroyed before DOM ready - skipping localStorage injection');
        return;
      }
```

a7m-1st: Why don't we use win.webContents.once here as well?

a7m-1st:
A small note that the `initState` == done only means that it transitioned between those two. So technically if you are at home page, or this last installation step page then `initState` is done:
Screenshot 2025-10-02 105400

To
Screenshot 2025-10-02 105411

electron/main/index.ts
```diff
@@ -919,14 +936,153 @@ async function createWindow() {
...
+  const uvExists = await isBinaryExists('uv');
+  const bunExists = await isBinaryExists('bun');
+
+  // Check if installation was previously completed
+  const backendPath = getBackendPath();
+  const installedLockPath = path.join(backendPath, 'uv_installed.lock');
+  const installationCompleted = fs.existsSync(installedLockPath);
+
+  const needsInstallation = !versionExists || savedVersion !== currentVersion || !uvExists || !bunExists || !installationCompleted;
+
```
a7m-1st: I think we can also check for the env path. Currently if env is removed, the result is:
```sh
[info] Installation check result: {
  needsInstallation: false,
  versionExists: true,
  versionMatch: true,
  uvExists: true,
  bunExists: true,
  installationCompleted: true
}
```

electron/main/install-deps.ts
```diff
@@ -315,20 +350,32 @@ export async function installDependencies(): Promise<PromiseReturnType> {
...
-    const installSuccess = await runInstall([])
+    const installSuccess = await runInstall([], version)
     if (installSuccess.success) {
         handleInstallOperations.spawnBabel()
+
+        // Clean up old venvs after successful installation
+        log.info('[DEPS INSTALL] Cleaning up old virtual environments...')
+        await cleanupOldVenvs(version)
+        log.info('[DEPS INSTALL] Old venvs cleanup completed')
+
```
a7m-1st, line 558: @Wendong-Fan would this remove old dependencies only after version update?

electron/main/index.ts
```diff
@@ -919,14 +936,153 @@ async function createWindow() {
...
+                console.log('[ELECTRON] initState updated to done, reloading page...');
+                return true; // Signal that we need to reload
+              }
+            }
+            return false; // No reload needed
+          } catch (e) {
+            console.error('[ELECTRON] Failed to update initState:', e);
+            return false;
+          }
+        })();
```
a7m-1st, line 1070: I am not sure if this caused it, but I was logged out at one point. I saw some kind of localStorage error. But couldn't replicate it.

a7m-1st (commented):
Thanks @Wendong-Fan for the update. Honestly I was thinking to always run "uv sync" by default instead of checking the versionFile only.

Or perhaps now since we can manage env per version, we can just add that check to `checkAndInstallDepsOnUpdate`. Bcz honestly the frontend seems fine if we just made sure it can detect the 'uvcorn' installation, so maybe we don't need the localStorage check or the frontend updates as things might get little more complicated.

Would love to hear your opinion though.

a7m-1st:
for reference:
image

a7m-1st:
@Wendong-Fan 
I will be making tests for those functions to make sure all problems are tracked.

a7m-1st:
Hi there @Wendong-Fan , its not perfect but I opened a PR to this branch to add tests https://github.com/eigent-ai/eigent/pull/449

electron/main/index.ts
Wendong-Fan: updated

PR #454, opened 2025-10-04
enhance: browser_install PR451

Wendong-Fan:
1.  The original code used `lib/site-packages`, but Python venvs actually use `lib/python3.X/site-packages` where X is the version number, the toolkit directory could never be found. fixed it to automatically detect the Python version and build the correct path.

  2. The code assumed stdout and stderr are always available, but they can be null in certain scenarios, added safety checks before trying to read from these streams to prevent crashes.

  3. When testing if npm exists, the code only listened for the 'close' event. If npm doesn't exist or fails to run, the 'error'
  event fires instead, causing the check to hang forever. added an error handler to properly handle this case.

  4. the npm/build/playwright processes only had 'close' handlers but no 'error' handlers. If a process fails to start,
  it triggers 'error' not 'close', which would cause the installation to hang. added error handlers for all three processes.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/install-deps.ts
+++ b/electron/main/install-deps.ts
@@ -333,11 +333,28 @@ export async function installDependencies(version: string): Promise<PromiseRetur
     installHybridBrowserDependencies: async (): Promise<boolean> => {
       try {
         // Find the hybrid_browser_toolkit ts directory in the virtual environment
-        const sitePackagesPath = path.join(venvPath, 'lib', 'site-packages');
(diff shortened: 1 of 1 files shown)


nitpicker55555 (approved):
Added isolated npm cache directory because users who previously used sudo npm may have permission conflicts with the system npm cache.

Issue #484, opened 2025-10-10
[Feature Request] Refactor toolkit decorators to use automated wrapping approach


Wendong-Fan:
### Motivation

Currently, when wrapping CAMEL toolkits, we need to manually override each method and apply the @listen_toolkit decorator. This approach is verbose and hard to maintain.

  Current implementation example (note_taking_toolkit.py):
```
  class NoteTakingToolkit(BaseNoteTakingToolkit, AbstractToolkit):
      @listen_toolkit(BaseNoteTakingToolkit.append_note)
      def append_note(self, note_name: str, content: str) -> str:
          return super().append_note(note_name=note_name, content=content)

      @listen_toolkit(BaseNoteTakingToolkit.read_note)
      def read_note(self, note_name: Optional[str] = "all_notes") -> str:
          return super().read_note(note_name=note_name)

      @listen_toolkit(BaseNoteTakingToolkit.create_note)
      def create_note(self, note_name: str, content: str, overwrite: bool = False) -> str:
          return super().create_note(note_name=note_name, content=content, overwrite=overwrite)

      @listen_toolkit(BaseNoteTakingToolkit.list_note)
      def list_note(self) -> str:
          return super().list_note()
```

  Issues with current approach:
  - Boilerplate code for each method
  - Easy to forget decorating new methods
  - Maintenance burden when base toolkit changes
  - Repetitive across all toolkit wrappers

  Proposed Solution

  Implement an automated decorator wrapping mechanism like below:

```
  @auto_listen_toolkit(BaseNoteTakingToolkit)
  class NoteTakingToolkit(BaseNoteTakingToolkit, AbstractToolkit):
      agent_name: str = Agents.document_agent
      # Methods automatically wrapped - no manual overrides needed
```

  Scope

  This enhancement should be applied to all CAMEL toolkit wrappers in the codebase, not just NoteTakingToolkit.

PR #486, opened 2025-10-10
enhance: wrapper listener

nitpicker55555:
#484 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/fileReader.ts
+++ b/electron/main/fileReader.ts
@@ -559,6 +559,28 @@ export class FileReader {
 			return [];
 		}
 	}
+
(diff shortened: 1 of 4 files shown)


nitpicker55555:
> I think it would be better to exclude helper methods like get_tools from being wrapped by auto_listen_toolkit. These methods are usually used for tool enumeration or initialization and don’t need event listening. Including them can trigger unnecessary activate/deactivate events or even cause startup errors when no task lock exists.
> 
> Filtering them out should make the wrapping logic cleaner and safer.

Thanks @fengju0213 ! fixed

PR #494, opened 2025-10-12
chore: update Google calendar

fengju0213:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Pakchoioioi:
https://github.com/user-attachments/assets/19269640-f86f-40a6-b5a1-8d83c8a037b0

there is a lot of bug here!
1. install then jump to 404
2. i need to input credential many times and the setting window didn't close when i install successfully ( but i am not sure if it was installed successfully)

Pakchoioioi:
5b2bbc8e-2454-4531-8318-c804234b4fb7
The issue that Google Calendar cannot retrieve the correct time for the next task is still unresolved.

nitpicker55555:
I think we need change  datetime.date.today() to  datetime.datetime.now() to give exact time

fengju0213:
> 10.14.2.1.mp4 
> there is a lot of bug here!
> 
> 1. install then jump to 404
> 2. i need to input credential many times and the setting window didn't close when i install successfully ( but i am not sure if it was installed successfully)

1.It's because the parameters are incorrect.

> I think we need change datetime.date.today() to datetime.datetime.now() to give exact time

updated  accurate to the hour ，directly use datetime.datetime.now() will break kv cache

fengju0213:
> 5b2bbc8e-2454-4531-8318-c804234b4fb7 The issue that Google Calendar cannot retrieve the correct time for the next task is still unresolved.

updated

backend/app/utils/oauth_state_manager.py
```diff
@@ -0,0 +1,109 @@
...
+            if provider in self._states:
+                state = self._states[provider]
+                state.status = status
+                state.error = error
+                state.result = result
+                if status in ["success", "failed", "cancelled"]:
+                    state.completed_at = datetime.now()
+                logger.info(f"Updated {provider} OAuth status to {status}")
+    
+    def cleanup_old_states(self, max_age_seconds: int = 3600):
```
waleedalzarooni: I don't think `cleanup_old_states` is called anywhere in the codebase

Pakchoioioi:
Google Calendar configuration failed @fengju0213 

https://github.com/user-attachments/assets/1922e246-6e2e-4d97-81e2-cc9bc3c877e3

backend/app/utils/oauth_state_manager.py
fengju0213: thanks @waleedalzarooni  fixed

fengju0213:
> Google Calendar configuration failed @fengju0213
> 
>  google.calendar.mp4

fixed can you try it again? the https://github.com/eigent-ai/eigent/issues/523 also be fixed

Pakchoioioi:
1、After the web page displays "Authorization successful!," is it possible to directly launch the client application? (Instead of just staying on the success page.)
2、This appears to be an issue with Google Calendar integration. The tool does not show as installed even after successful authentication, whether adding it via the "Add Worker" section or the tool settings page.
3、Can the authentication speed be improved? Currently, the verification process takes 3-5 seconds before redirecting to the authorization page.

PR #511, opened 2025-10-15
chore: tool_schema_fix

fengju0213:
### Description
use deepcopy temporary fix tool schema and ehance notion

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/utils/agent.py
+++ b/backend/app/utils/agent.py
@@ -323,6 +323,17 @@ class ListenChatAgent(ChatAgent):
                 else:
                     result = raw_result
                     mask_flag = False
+                # Prepare result message with truncation
+                if isinstance(result, str):
+                    result_msg = result
+                else:
+                    result_str = repr(result)
+                    MAX_RESULT_LENGTH = 500
+                    if len(result_str) > MAX_RESULT_LENGTH:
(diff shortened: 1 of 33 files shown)


backend/app/utils/toolkit/notion_mcp_toolkit.py
```diff
@@ -6,6 +6,25 @@
...
+        """
+        function_info = schema.get("function", {})
+        function_name = function_info.get("name", "")
+        parameters = function_info.get("parameters", {})
+        properties = parameters.get("properties", {})
+        required = parameters.get("required", [])
+        
+        # Modify the notion-create-pages function to make parent optional
+        if function_name == "notion-create-pages":
+            required.remove("parent")
```
nitpicker55555, line 23: should we add  if "parent" in required ?

fengju0213:
already found the root case, the model platform output by the front end is wrong（Gemini）
gemini model will pop strcit
image

backend/app/utils/toolkit/notion_mcp_toolkit.py
fengju0213: thanks @nitpicker55555  updated

Issue #469, opened 2025-10-06
[Feature Request] Adding project_id field to database server schema


a7m-1st:
### Motivation

To support #415 we would need to render multiple tasks per page, thus requiring us to save each Chat with project_id field as well.

Currently I identified `ChatHistory` model to need this change as we will be saving the changes through its controller. 

Would like to ask if I updated the model:
1. Will the database migration happen on cicd?
2. Do I need to worry about old projects or will it just return null? For now I suggest we set `project_id == task_id` for older user projects if possible.
3. Just updating ChatHistory, ChatHistoryIn & ChatHistoryOut is enough?

cc: @Wendong-Fan @Pakchoioioi 

### Solution

```python
class ChatHistory(AbstractModel, DefaultTimes, table=True):
    id: int = Field(default=None, primary_key=True)
    user_id: int = Field(index=True)
    task_id: str = Field(index=True, unique=True)
    project_id: str = Field(index=True, unique=True) # The new field
    question: str
    language: str
    model_platform: str
    model_type: str
    api_key: str
    api_url: str = Field(sa_column=Column(String(500)))
    max_retries: int = Field(default=3)
    file_save_path: Optional[str] = None
    installed_mcp: str = Field(sa_type=JSON, default={})
    project_name: str = Field(default="", sa_column=Column(String(128)))
    summary: str = Field(default="", sa_column=Column(String(1024)))
    tokens: int = Field(default=0, sa_column=(Column(Integer, server_default="0")))
    spend: float = Field(default=0, sa_column=(Column(Float, server_default="0")))
    status: int = Field(default=1, sa_column=Column(ChoiceType(ChatStatus, SmallInteger())))
```

```python
op.create_table(
      "chat_history",
      sa.Column("deleted_at", sa.DateTime(), nullable=True),
      sa.Column("created_at", sa.TIMESTAMP(), server_default=sa.text("CURRENT_TIMESTAMP"), nullable=True),
      sa.Column("updated_at", sa.TIMESTAMP(), server_default=sa.text("CURRENT_TIMESTAMP"), nullable=True),
      sa.Column("id", sa.Integer(), nullable=False),
      sa.Column("user_id", sa.Integer(), nullable=False),
      sa.Column("task_id", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
      sa.Column("project_id", sqlmodel.sql.sqltypes.AutoString(), nullable=False), # New field
[...]

PR #512, opened 2025-10-15
update chat history model: add project id

weer0026:
### Description

#469 Finish this isuess's job

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/utils/agent.py
+++ b/backend/app/utils/agent.py
@@ -323,6 +323,17 @@ class ListenChatAgent(ChatAgent):
                 else:
                     result = raw_result
                     mask_flag = False
+                # Prepare result message with truncation
+                if isinstance(result, str):
+                    result_msg = result
+                else:
+                    result_str = repr(result)
+                    MAX_RESULT_LENGTH = 500
+                    if len(result_str) > MAX_RESULT_LENGTH:
+                        result_msg = result_str[:MAX_RESULT_LENGTH] + f"... (truncated, total length: {len(result_str)} chars)"
+                    else:
+                        result_msg = result_str
+
                 asyncio.create_task(
                     task_lock.put_queue(
                         ActionDeactivateToolkitData(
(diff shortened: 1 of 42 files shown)


a7m-1st:
Thanks @weer0026 this would simplify a ton.

server/alembic/versions/2025_10_15_1446-eec7242b3a9b_modify_chat_history_add_project_id.py
```diff
@@ -0,0 +1,36 @@
...
+down_revision: Union[str, None] = "d74ab2a44600"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("chat_history", sa.Column("project_id", sqlmodel.sql.sqltypes.AutoString(), nullable=True))
+    op.create_index(op.f("ix_chat_history_project_id"), "chat_history", ["project_id"], unique=True)
```
Wendong-Fan, line 27: should we set `unique=False` here? i think the task id should be unique, but we can allow duplicated project id here

server/app/model/chat/chat_history.py
```diff
@@ -16,6 +16,7 @@ class ChatHistory(AbstractModel, DefaultTimes, table=True):
     id: int = Field(default=None, primary_key=True)
     user_id: int = Field(index=True)
     task_id: str = Field(index=True, unique=True)
+    project_id: str = Field(index=True, unique=True, nullable=True)
```
Wendong-Fan, line 19: same for here

server/alembic/versions/2025_10_15_1446-eec7242b3a9b_modify_chat_history_add_project_id.py
a7m-1st: Yes thats right, it will be a flat hierarchy of unique tasks associated to one project_id.
Thus project_id will have many task_ids

PR #519, opened 2025-10-16
fix: correct ordered list rendering in Markdown where all items appea…

### LINKED ISSUE #515: [BUG] Markdown ordered list always renders as 1.

0.0.72 beta

### System information

mac m chip

### Problem description

Markdown ordered list always renders as 1.
When using numbered lists in Markdown, all list items are rendered as 1. instead of incrementing numbers. For example:

Image

### log

Image

### Additional context

_No response_

FooFindBar:
…red as "1."

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

close #515


FooFindBar:
> Good fix direction, but there's an issue:
> 
> The `olPadding` prop is used in `src/components/WorkFlow/node.tsx` (3 places with `olPadding="pl-0"`). Commenting out the `ol` component breaks this dynamic padding control.
> 
> **Suggestion:** Keep the `ol` component, just change `list-inside` to `list-outside`:
> 
> ```tsx
> ol: ({ children }) => (
>   <ol
>     className={`list-decimal list-outside text-[13px] text-primary mb-2 ${olPadding}`}
>     style={{ paddingLeft: olPadding || '1rem' }}
>   >
>     {children}
>   </ol>
> ),
> ```
> 
> This fixes the numbering issue without breaking existing functionality.

image
image

Thanks for the suggestion!
I tried using the text extracted from the image in the MarkDown component for testing, but the ordered list numbers were still disconnected.
That’s why I used the current approach to fix the issue.

Pakchoioioi:
Screenshot 2025-10-17 at 12 14 47

Issue #501, opened 2025-10-14
[BUG] Agent assign_task singal lost


nitpicker55555:
### What version of eigent are you using?

0.0.72

### System information

mac m4

### Problem description

<img width="2350" height="1140" alt="Image" src="https://github.com/user-attachments/assets/c8ebbe18-4201-4fc4-aab7-5c2f4ab35986" />

<img width="1386" height="810" alt="Image" src="https://github.com/user-attachments/assets/6bffaf9f-4582-48f6-8a69-ab7ab565b4d9" />

Only document agent has received assign_task signal 
search agent has not recevied assign_task signal so agent in interface has not been focused

### log

[eigent-0.0.72-darwin-arm64-1760391714698.log](https://github.com/user-attachments/files/22901193/eigent-0.0.72-darwin-arm64-1760391714698.log)

nitpicker55555:
in  eigent/backend/app/utils/workforce.py
task_lock.put_queue in _find_assignee() stop()  has not been await, but has await in other cases like _post_task()

nitpicker55555:
[workforce_debug_2025-10-13.log](https://github.com/user-attachments/files/22945614/workforce_debug_2025-10-13.log)

2025-10-13 22:37:55.990 | DEBUG    | app.utils.workforce:eigent_start:107 - start the workforce subtasks=[Task(id='1760391444331-7092.1', content='Search for the latest news articles related to Munich using current online news platforms. Summarize the top three news stories in bullet points, providing the title, date, and a brief summary (2-3 sentences) for each story. The output should be structured as a markdown bullet list for clarity.', state='FAILED')]
2025-10-13 22:37:57.918 | DEBUG    | app.utils.workforce:_find_assignee:129 - [WF] ASSIGN 1760391444331-7092.1 -> 777b6691-2d95-41ae-9a56-2e21a390eadf deps=['1760391444331-7092']
2025-10-13 22:37:57.919 | DEBUG    | app.utils.workforce:_find_assignee:129 - [WF] ASSIGN 1760391444331-7092 -> 93b12714-48e1-4cb3-a11d-fa8aaa9c7a54 deps=[]
2025-10-13 22:37:57.919 | DEBUG    | app.utils.workforce:_post_task:161 - [WF] POST  1760391444331-7092 -> 93b12714-48e1-4cb3-a11d-fa8aaa9c7a54

nitpicker55555:
solved by #525

PR #525, opened 2025-10-17
chore: remove main task in pending_task 

fengju0213:
### Fix: Exclude root task from assignment; assign only executable subtasks

- Summary
  - Previously, the product enqueued the root task into the pending queue, so it was eligible for assignment and posting. This diverged from CAMEL’s default behavior where only executable subtasks are assigned, and the root task acts as an orchestrator.
  - This PR aligns our behavior with CAMEL by not enqueuing the root task. Only subtasks are now queued and assigned.

- What’s changed
  - Removed enqueuing of the root task in `backend/app/utils/workforce.py` (`eigent_make_sub_tasks`): the line that appended the root task to `_pending_tasks` was deleted.
  - No changes to downstream assignment logic; we continue to call `super()._find_assignee` and `super()._post_task`, but the input task set now excludes the root task.

- Why
  - Root task is an orchestration/goal node, not an executable unit.
  - Prevents accidental assignment, posting, timing, and retries of the root task.
  - Keeps logs/metrics clean and aligned with actual execution.
  - Matches CAMEL’s semantics, reducing behavioral drift.

- Behavior before
  - Root task was added to `_pending_tasks`, became assignable, and appeared in ASSIGN/POST logs.
  - Backend suppressed frontend notifications for the root task, so the UI rarely showed it, but logs still contained noise.

- Behavior after
  - Only subtasks are enqueued, assigned, and posted.
  - Root task remains as an anchor for planning/progress aggregation; it is not scheduled or executed.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -67,22 +67,25 @@ export default defineConfig(({ command, mode }) => {
         renderer: {},
       }),
     ],
-    server: process.env.VSCODE_DEBUG && (() => {
-      const url = new URL(pkg.debug.env.VITE_DEV_SERVER_URL)
-      return {
-        host: url.hostname,
-        port: +url.port,
-        proxy: {
-          '/api': {
-            target: env.VITE_PROXY_URL,
-            changeOrigin: true,
-            // rewrite: path => path.replace(/^\/api/, ''),
+    server: {
+      open: false,
+      ...(process.env.VSCODE_DEBUG && (() => {
+        const url = new URL(pkg.debug.env.VITE_DEV_SERVER_URL)
+        return {
+          host: url.hostname,
+          port: +url.port,
+          proxy: {
+            '/api': {
+              target: env.VITE_PROXY_URL,
+              changeOrigin: true,
+              // rewrite: path => path.replace(/^\/api/, ''),
+            },
           },
-        },
-      }
-    })(),
-    clearScreen: false,
+        }
+      })()),
+      clearScreen: false,
 
+    }
   }
 })
 


fengju0213:
related to https://github.com/eigent-ai/eigent/issues/501

backend/app/utils/workforce.py
```diff
@@ -85,7 +85,6 @@ def eigent_make_sub_tasks(self, task: Task):
         self.set_channel(TaskChannel())
         self._state = WorkforceState.RUNNING
         task.state = TaskState.OPEN
-        self._pending_tasks.append(task)
```
a7m-1st, line 88: Good point, but unless I missed some PRs in Camel side actually, we have extended the use of _pending_tasks queue to also host independent tasks rather have a separate queue. I also made it so that the main task goes to completed task when complete. 

Image

Reference: https://github.com/camel-ai/camel/pull/3238

The main thing to note is that if task is completed it passes through `task_state` event when completed which triggers frontend events:

### Extended function from /backend/app/workforce.py
```python
async def _handle_completed_task(self, task: Task) -> None:
        # DEBUG ▶ Task completed
        logger.debug(f"[WF] DONE  {task.id}")
        task_lock = get_task_lock(self.api_task_id)

        task_data = {
            "task_id": task.id,
            "content": task.content,
            "state": task.state,
            "result": task.result or "",
            "failure_count": task.failure_count,
        }
        
        await task_lock.put_queue(
            ActionTaskStateData(
                data=task_data
            )
        )

        return await super()._handle_completed_task(task)
```

a7m-1st:
> Root task was added to _pending_tasks, became assignable, and appeared in ASSIGN/POST logs.

The solution to this would be how I am implemented in Camel Side, just remove immediately after task decomposition through this function.

```python
# Decompose the task and append subtasks to _pending_tasks
await self.handle_decompose_append_task(
    next_task, reset=False
)

# Mark the main task as completed and remove from pending
await self._handle_completed_task(next_task)
logger.info(
    f"Main task {next_task.id} decomposed after "
    f"skip request"
)
```

fengju0213:
thanks for reviewing @a7m-1st ! I think we might be talking about slightly different layers here.

My change targets the *root task* (the prompt user submit), 
which shouldn't enter `_pending_tasks` or appear in ASSIGN/POST logs at all — it’s 
never meant to be executable or assigned.

What you mentioned about independent main tasks coexisting in `_pending_tasks` 
makes sense for real executable units, but that doesn’t conflict with this PR — 
root tasks remain orchestration-only, while subtasks and independent tasks 
are still enqueued as before.

a7m-1st:
Thanks @fengju0213 . All right then I got it. I was just concerned that the root task might be sent over for SSE State changes but after testing I didn't find any discrepancies to both the chatStore.task & SSE network requests 👍.

But what I was thinking is that we can `append -> decompose -> remove from pending_tasks & send to _completed_tasks queue -> continue eigent`. Actually the _handle_compeleted_task specifically checks for the below among other things:

```python
# Handle parent task completion logic
parent = task.parent
if parent:
    # Check if all subtasks are completed and successful
    all_subtasks_done = all(
        any(
            t.id == sub.id and t.state == TaskState.DONE
            for t in self._completed_tasks
        )
        for sub in parent.subtasks
    )
    if all_subtasks_done:
        # Collect results from successful subtasks only
        successful_results = []
        for sub in parent.subtasks:
            completed_subtask = next(
                (
                    t
                    for t in self._completed_tasks
                    if t.id == sub.id and t.state == TaskState.DONE
                ),
                None,
            )
            if completed_subtask and completed_subtask.result:
                successful_results.append(
                    f"--- Subtask {sub.id} Result ---\n"
                    f"{completed_subtask.result}"
                )

        # Set parent task state and result
        parent.state = TaskState.DONE
        parent.result = (
            "\n\n".join(successful_results)
            if successful_results
            else "All subtasks completed"
        )

        logger.debug(
            f"All subtasks of {parent.id} are done. "
            f"Marking parent as complete."
        )
        # Treat the parent task as a completed task to unblock
        # its dependents. Since it was never sent to a worker,
        # we call this method recursively.
        await self._handle_completed_task(parent)
```

So if main task is none existing in _pending_tasks then this above logic would be pointless. Unless I am misunderstanding the use.

a7m-1st:
TLDR: My suggestion I know it might be inefficient, but keeps the essence, where any updates / error handling to the root task would be handled too. @fengju0213 

```python
self._pending_tasks.append(task)

# Decompose the task into subtasks first
subtasks_result = self._decompose_task(task)

# Remove from pending_tasks and append to completed tasks
self._handle_completed_task(task)
```

Edit: Or at least:
```python
self._completed_tasks.append(task)

# Decompose the task into subtasks first
subtasks_result = self._decompose_task(task)
```

fengju0213:
> TLDR: My suggestion I know it might be inefficient, but keeps the essence, where any updates / error handling to the root task would be handled too. @fengju0213
> 
> ```python
> self._pending_tasks.append(task)
> 
> # Decompose the task into subtasks first
> subtasks_result = self._decompose_task(task)
> 
> # Remove from pending_tasks and append to completed tasks
> self._handle_completed_task(task)
> ```
> 
> Edit: Or at least:
> 
> ```python
> self._completed_tasks.append(task)
> 
> # Decompose the task into subtasks first
> subtasks_result = self._decompose_task(task)
> ```

This change simply excludes the root task from pending/assignment, matching CAMEL’s semantics.and i think parent completion is still handled when all subtasks are DONE via handle_completed_task？

a7m-1st:
Should be okay for now as I don't see affecting anything. Just following similar pattern in multi-turn PR. 

Thanks for your consideration @fengju0213 !

a7m-1st (approved):

PR #526, opened 2025-10-17
chore: add time out for question_confirm

fengju0213:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/service/chat_service.py
+++ b/backend/app/service/chat_service.py
@@ -88,21 +88,11 @@ async def step_solve(options: Chat, request: Request, task_lock: TaskLock):
                     assert isinstance(item, ActionImproveData)
                     question = item.data
                 if len(question) < 12 and len(options.attaches) == 0:
-                    try:
-                        confirm = await asyncio.wait_for(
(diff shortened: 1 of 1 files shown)


backend/app/service/chat_service.py
```diff
@@ -88,7 +88,21 @@ async def step_solve(options: Chat, request: Request, task_lock: TaskLock):
...
+                        logger.warning(f"question_confirm timeout for task {options.task_id}")
+                        # Notify frontend to avoid appearing stuck and remain in simple-answer path
+                        yield sse_json(
+                            "notice",
+                            {"notice": "Quick classification timed out. Responding directly."},
+                        )
+                        confirm = sse_json(
+                            "wait_confirm",
+                            {"content": "Classification timed out. Please confirm or try again."},
+                        )
```
a7m-1st: Thanks for the PR.
I think it will be cleaner to add this to the question confirm function @fengju0213

Wendong-Fan: here would block the running, it's not what we want, we need to set `confirm = True` here

PR #527, opened 2025-10-17
fix: preserve user login session across app restarts

FooFindBar:

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/webview.ts
+++ b/electron/main/webview.ts
@@ -271,7 +271,7 @@ export class WebViewManager {
         webViewInfo.view.webContents.removeAllListeners()
         webViewInfo.view.webContents.session.clearCache()
         webViewInfo.view.webContents.session.clearStorageData({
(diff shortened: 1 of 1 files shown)


electron/main/webview.ts
```diff
@@ -271,7 +271,7 @@ export class WebViewManager {
         webViewInfo.view.webContents.removeAllListeners()
         webViewInfo.view.webContents.session.clearCache()
         webViewInfo.view.webContents.session.clearStorageData({
-          storages: ['cookies', 'localstorage', 'websql', 'indexdb', 'serviceworkers', 'cachestorage']
+          storages: ['cookies', 'websql', 'indexdb', 'serviceworkers', 'cachestorage']
```
a7m-1st, line 274: Thanks for the PR @FooFindBar, just a light concern:
- Since we are not clearing on destroy; On logout is localStorage cleared?

PR #537, opened 2025-10-20
fix: prevent API call before backend is fully started

### LINKED ISSUE #529: [BUG] Task submission fails if sent immediately after dependency installation

0.0.72

### System information

win10

### Problem description

When the user installs dependencies and immediately submits a task, the request may fail because the backend service is not fully started yet.

### log

null

### Additional context

_No response_

FooFindBar:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

close #529

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -16,6 +16,7 @@ import { copyBrowserData } from './copy'
 import { findAvailablePort } from './init'
 import kill from 'tree-kill';
 import { zipFolder } from './utils/log'
+import archiver from 'archiver';
 import axios from 'axios';
 import FormData from 'form-data';
 import { checkAndInstallDepsOnUpdate, PromiseReturnType, getInstallationStatus } from './install-deps'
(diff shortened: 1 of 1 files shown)


electron/main/index.ts
```diff
@@ -1186,15 +1256,14 @@ const checkAndStartBackend = async () => {
     if (isToolInstalled.success) {
       log.info('Tool installed, starting backend service...');
 
-      // Notify frontend installation success
```
fengju0213, line 1189: Thanks for the fix! a question though:
Looking at `startBackend()` in `init.ts:209`, it resolves as soon as it sees `"Uvicorn running on"`:

```typescript
if (!started && data.toString().includes("Uvicorn running on")) {
    started = true;
    resolve(node_process);  // Resolves immediately
}
```
There could still be a gap between Uvicorn starting and FastAPI routes being fully loaded. If users submit tasks immediately after installation completes, won't we still hit the race condition from issue #529?

fengju0213: Should we add a health check poll before notifying the frontend?

PR #549, opened 2025-10-21
feat: multi OS rm -rf

### LINKED ISSUE #548: [BUG] package.json tries to run unsupported command in Windows platforms

0.0.72

### System information

3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)] win32

### Problem description

- there is a command which tries to run `rm -rf` which is not supported in windows
Image

### log

Image

### Additional context

_No response_

a7m-1st:
### Description
fixes #548 
Also refixes #504 

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/service/chat_service.py
+++ b/backend/app/service/chat_service.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import json
 from pathlib import Path
 import platform
 from typing import Literal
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,8 +14,8 @@
   "type": "module",
   "scripts": {
     "compile-babel": "cd backend && uv run pybabel compile -d lang",
-    "clean-cache": "rm -rf node_modules/.vite",
-    "dev": "npm run clean-cache && npm run compile-babel && vite",
```
Wendong-Fan, line 18: i think we need to keep `npm run compile-babel`

a7m-1st: All right, but the issue with that is we are using a different cache path, we will have /backend/.venv (which won't be used by eigent) and another one at ~.eigent/backend/venvs/... :
```
this.node_process = spawn(uv_path, [
        'sync',
        '--no-dev',
        '--cache-dir', getCachePath('uv_cache'),
        ...extraArgs], {
        cwd: backendPath,
        env: {
            ...process.env,
            UV_TOOL_DIR: getCachePath('uv_tool'),
            UV_PYTHON_INSTALL_DIR: getCachePath('uv_python'),
            UV_PROJECT_ENVIRONMENT: venvPath,
        }
    })
```

Issue #573, opened 2025-10-23
[BUG] developer agent got stuck


yuqinxie-sketch:
### What version of eigent are you using?

0.0.72

### System information

mac m chip

### Problem description

developer agent got stuck, used claude sonnet 4.5 model, the task prompt is: Get all the 3116 Stargazers' information of https://github.com/unitycatalog/unitycatalog Save into a Csv file including their linkedin and x links which is usually present in their github profiles.

### log

[eigent-0.0.72-darwin-arm64-1761239070154.log](https://github.com/user-attachments/files/23103439/eigent-0.0.72-darwin-arm64-1761239070154.log)

Pakchoioioi:
will config with CAMEL next version

Issue #488, opened 2025-10-10
[BUG] Unexpected Installation page


Saedbhati:
### What version of eigent are you using?

0.0.72

### System information

window 11

### Problem description

Resolved, Downloaded, etc. is being treated as an installer running, so `detectInstallationLogs `creates `uv_installing.lock` and notifies the UI even when there was no explicit install run.

https://github.com/user-attachments/assets/e749235b-94de-4b50-b88b-1cd29b68909c

### log

[eigent-0.0.72-win32-x64-1760095917691.log](https://github.com/user-attachments/files/22848977/eigent-0.0.72-win32-x64-1760095917691.log)

a7m-1st:
Good catch.. actually we can remove auto detection feature from the logs:

```js
/** intall-deps.ts */
  displayFilteredLogs(data:String) {
      if (!data) return;
      const msg = data.toString().trimEnd();

      //Remove
      //Detect if uv sync is run
      detectInstallationLogs(msg);
     //To Remove

      if (msg.toLowerCase().includes("error") || msg.toLowerCase().includes("traceback")) {
          log.error(`BACKEND: [DEPS INSTALL] ${msg}`);
          safeMainWindowSend('install-dependencies-log', { type: 'stderr', data: data.toString() });
      } else {
          log.info(`BACKEND: [DEPS INSTALL] ${msg}`);
          safeMainWindowSend('install-dependencies-log', { type: 'stdout', data: data.toString() });
      }
  }
```

Fixes issue instantly, but means that we will rely that the installation is always triggered on main flow correctly. What do you think @Wendong-Fan ?

PR #581, opened 2025-10-24
fix: remove detect install

a7m-1st:
### Description
fixes - #573 is caused by "uvicorn" installation detection, thus triggering the whole uv cache clearing, npm installation etc...
possible cause for #488 

### Note:
@Wendong-Fan @nitpicker55555 according to the design it shouldn't have been triggered bcz its on separate channel all together from the main uv.

But generally I think after Wendong's fix to isolate, it should be stable. I.e. hopefully installation won't be triggered on runtime. What do you think?

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Wendong-Fan:
hey @a7m-1st , should we close this PR as the isolate env feature has been merged?

a7m-1st:
If the issue is not appearing anymore, then perhaps we can close this @Wendong-Fan .

Issue #595, opened 2025-10-31
[BUG] Clicking “Close” button on the login status page has no effect


Pakchoioioi:
### What version of eigent are you using?

0.0.71

### System information

mac & windows

### Problem description

Clicking “Close” button on the login status page has no effect

### log

<img width="1313" height="813" alt="Image" src="https://github.com/user-attachments/assets/855dc0c6-f30f-4413-9ea8-1ebc684e9343" />

PR #596, opened 2025-10-31
enhance: add close button to login page

FooFindBar:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

#595

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -582,7 +582,6 @@ function registerIpcHandlers() {
 
   // ==================== window control handler ====================
   ipcMain.on('window-close', (_, data) => {
-    console.log('window-close', data);
     if(data.isForceQuit) {
       return app?.quit()
     }
--- a/src/components/WindowControls/index.tsx
+++ b/src/components/WindowControls/index.tsx
@@ -45,8 +45,9 @@ export default function WindowControls() {
 				onClick={(e) => {
 					e.stopPropagation();
 					e.preventDefault();
-					// Use forceQuit=true to bypass before-close handler
-					window.electronAPI.closeWindow(true);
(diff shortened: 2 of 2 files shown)


Douglasymlai:
There are some design issue needed to be fixed

1. Header Issue
The Eigent name should not appear in the header.
Screenshot 2025-11-04 at 11 09 31

2. Home Icon Change
The home icon has been switched to a grid icon.
I’m not sure if this change was intentional or a mistake — please clarify the reason for this update.
Screenshot 2025-11-04 at 11 08 59

electron/main/index.ts
```diff
@@ -582,6 +582,7 @@ function registerIpcHandlers() {
 
   // ==================== window control handler ====================
   ipcMain.on('window-close', (_, data) => {
+    console.log('window-close', data);
```
Wendong-Fan: should this be cleaned?

src/components/WindowControls/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+			>
+				<Square className="w-4 h-4" />
+			</div>
+			<div
+				className="control-btn h-full flex-1"
+				onClick={(e) => {
+					e.stopPropagation();
+					e.preventDefault();
+					// Use forceQuit=true to bypass before-close handler
+					window.electronAPI.closeWindow(true);
```
Wendong-Fan: would use  `forceQuit=true` to bypass before-close handler be problematic and make the close logic hard to maintain? using `forceQuit=true` seems would not do any clean and would force quiet even there's project running

nitpicker55555: Fixed

nitpicker55555 (approved):

Issue #599, opened 2025-10-31
[Feature Request] Refactor route registration


Archie818:
### Motivation

- No visibility of route mappings: There's no single place to see all API endpoints.
- Difficult to locate quickly: It's unclear which file corresponds to /chat.
- Difficult to debug: When errors occur, it's hard to know where to start looking.

### Solution

Replace auto-discovery router registration with explicit router registry for better visibility and maintainability

PR #600, opened 2025-10-31
Fix refactor route registration issue

Archie818:
### Description

Replace auto-discovery router registration with explicit router registry for better visibility and maintainability

Issue: https://github.com/eigent-ai/eigent/issues/599

- [ ] Bug fix
- [X ] New Feature
- [ ] Documentation update
- [ ] Other


backend/app/router.py
```diff
@@ -0,0 +1,59 @@
...
+            "tags": ["tool"], 
+            "description": "Tool installation and management"
+        },
+    ]
+    
+    for config in routers_config:
+        app.include_router(
+            config["router"],
+            prefix=prefix,
+            tags=config["tags"]
```
Wendong-Fan, line 52: in the controller we have defined tags, would here introduce duplicated tag?

Wendong-Fan (commented):
thanks @Archie818 , left one comment below

backend/app/router.py
Archie818: @Wendong-Fan This is a good question. To maintain consistency of tags, I suggest managing the tags for a controller in one place, within the router.
I will changed the relevant code later.

Wendong-Fan (approved):

Issue #577, opened 2025-10-23
[Feature Request] Duplicate task boxes appear after responding with human_toolkit in multi-turn conversation


Pakchoioioi:
### Motivation

There’s an issue on the frontend for multi-turn conversations — after I respond with human_toolkit, two task boxes appear instead of one.

<img width="3342" height="1814" alt="Image" src="https://github.com/user-attachments/assets/cc7c6207-b74a-4dbe-a1b5-bda28216048e" />

a7m-1st:
This issue is very similar to what happens if you click "skip" when some tasks are running

Pakchoioioi:
this might related to message box components.

PR #602, opened 2025-11-03
fix: prevent duplicate task boxes when responding with human_toolkit …

FooFindBar:
…in multi-turn conversations (#577)

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


utils/__pycache__/traceroot_wrapper.cpython-310.pyc
fengju0213: i think this pr shouldn't be submit？

PR #614, opened 2025-11-06
feat(chat_controller): add SSE timeout handling and improve logging

luoyou:
…tructure

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Wendong-Fan (commented):
added enhance PR: https://github.com/eigent-ai/eigent/pull/618, please help review

fengju0213:
We currently only emit an SSE “error” event and exit timeout_stream_wrapper, but never stop the underlying step_solve coroutine or clean up the task state.
As a result, there may be cases where the workforce and its background tasks continue running, and the TaskLock remains registered, causing the backend to keep consuming resources even after the client disconnects — leading to a potential resource leak.

Wendong-Fan (approved):

Issue #619, opened 2025-11-07
[BUG] When sending a task, two task boxes were created


Pakchoioioi:
### What version of eigent are you using?

0.0.72 st_calenfar_fix branch

### System information

in this branch: 
https://github.com/eigent-ai/eigent/pull/494

### Problem description

When sending a task, two task boxes were created and this task didn't run.
The to_sub_tasks step appears to have been triggered twice, resulting in duplicate requests.

<img width="1624" height="978" alt="Image" src="https://github.com/user-attachments/assets/8345dc4e-bc2b-46fa-9e9c-53eb489059e0" />

<img width="878" height="725" alt="Image" src="https://github.com/user-attachments/assets/4f018182-8e51-44e1-9c5e-71614c28239a" />

### log

data: {"step": "confirmed", "data": {"question": "帮我看下google calendar 下一个会议是什么时候，是什么会议，有哪些人参会"}}

data: {"step": "to_sub_tasks", "data": {"summary_task": "Task|帮我看下google calendar 下一个会议是什么时候，是什么会议，有哪些人参会", "sub_tasks": [{"id": "1762503273000-2985.1", "content": "Retrieve the details of the next scheduled event on the user's Google Calendar, including the start time, title of the meeting, and the list of all participants' names and email addresses. Format the result as a JSON object with the keys: 'start_time', 'meeting_title', and 'participants', where 'participants' is an array of objects each with 'name' and 'email'.", "state": "FAILED", "subtasks": []}]}}

data: {"step": "deactivate_agent", "data": {"agent_name": "question_confirm_agent", "agent_id": "3f9fff03-e95a-4ad6-a3d4-6bdd73b639c4", "process_task_id": "", "message": "yes", "tokens": 357}}

data: {"step": "create_agent", "data": {"agent_name": "task_agent", "agent_id": "17cef41b-5158-40e8-b1e2-0a88777dddb4", "tools": []}}

data: {"step": "create_agent", "data": {"agent_name": "search_agent", "agent_id": "c249ac7b-6745-4832-a5c1-3238f9176434", "tools": ["Search Toolkit", "Browser Toolkit", "Human Toolkit", "Note Taking Toolkit", "Terminal Toolkit"]}}

data: {"step": "deactivate_toolkit", "data": {"agent_name": "search_agent", "process_task_id": "", "toolkit_name": "Browser Toolkit", "method_name": "register agent", "message": "null"}}

[...]

PR #631, opened 2025-11-08
enhance: replay and fix related bugs

a7m-1st:
depends on #524 

### Description
- Replicated Prompt, tried in Calendar branch, replicated SSE in test cases but couldn't pinpoint issue #619. I added test cases to track the issue nonetheless.
- Can start task when replay is running i.e. can tolerate running Workforce while replay is running. This is possible when the replay task enters the onmessage loop, thus keeping the consistency of the `lockedStore` instances (take note this is because replay uses mock event source)

image

- add replay tag to projects
- Fix replay.ts bug where first question is not found & add typing.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


nitpicker55555:
hi can we also fix backend/app/utils/server/sync_step.py   TODO: revert to task_id to support multi-task project replay in this pr?

a7m-1st:
@nitpicker55555 for that one I am already working on it in `feat-migrate-api-to-projid` and it requires more than changing sync_step.
- Update file system
- Update history api
- Update history UI/UX (All 3 Components)

Wendong-Fan (approved):

PR #644, opened 2025-11-12
feat: grouped project History api

a7m-1st:
### Description
- Grouping tasks based on project id on client side is not viable due to pagination.

image

### Solution
- Group projects in backend with response format:

```json
"projects": [{
            "project_id": "1762426076155-1882",
            "project_name": "Task",
            "total_tokens": 38741,
            "task_count": 3,
            "latest_task_date": "2025-11-06T10:51:17.138116",
            "tasks": [
                {
                    "id": 4,
                    "task_id": "1762426085112-4289",
                    "project_id": "1762426076155-1882",
                    "question": "what can youdo?",
                    "language": "en",
                    "model_platform": "openai",
                    "model_type": "gpt-4.1",
                    "api_key": "",
                    "api_url": "cloud",
                    "max_retries": 3,
                    "file_save_path": "string",
                    "installed_mcp": "string",
                    "project_name": "Task",
                    "summary": "search again",
                    "tokens": 1328,
                    "status": 1,
                    "created_at": "2025-11-06T10:48:06.001684",
                    "updated_at": "2025-11-06T10:50:13.929855"
                },
                {
                    "id": 5,
                    "task_id": "1762426104975-6695",
                    "project_id": "1762426076155-1882",
                    "question": "what can youdo?",
                    "language": "en",
                    "model_platform": "openai",
                    "model_type": "gpt-4.1",
                    "api_key": "",
                    "api_url": "cloud",
                    "max_retries": 3,
                    "file_save_path": "string",
                    "installed_mcp": "string",
                    "project_name": "Task",
                    "summary": "serach the web",
                    "tokens": 12471,
                    "status": 1,
                    "created_at": "2025-11-06T10:49:36.530440",
                    "updated_at": "2025-11-06T10:51:53.243502"
                },
                {
                    "id": 6,
                    "task_id": "1762426276298-4324",
                    "project_id": "1762426076155-1882",
                    "question": "search again",
                    "language": "en",
                    "model_platform": "openai",
                    "model_type": "gpt-4.1",
                    "api_key": "",
                    "api_url": "cloud",
                    "max_retries": 3,
                    "file_save_path": "string",
                    "installed_mcp": "string",
                    "project_name": "Task",
                    "summary": "serach the web",
                    "tokens": 24942,
                    "status": 1,
                    "created_at": "2025-11-06T10:51:17.138116",
                    "updated_at": "2025-11-06T10:51:53.527306"
                }
            ],
            "total_completed_tasks": 0,
[...]

PR #645, opened 2025-11-13
feat: migrate api to projid

a7m-1st:
### All in all
- I would confidently say that this version of separating the tasks into separate replay channels would future proof the system because separation of concerns.
- And any issues with the channel, is now isolated to a single chatStore from top down.
`1 replay == 1 chatStore (1 workforce + Many quick replies)`
Previously: `1 replay == Many * chatStore (1 workforce + Many quick replies)`

### Description
- new `/grouped` History api endpoint (which falls back for client-side grouping & is backwards compatible)
- Task id is synced with sync_step via storing it with the func:
  a. new chat -> update current taskId
  b. Improve API -> update current taskId IFF workforce is enabled
```
    # Set the initial current_task_id in task_lock
    set_current_task_id(data.project_id, data.task_id)
```

### Behavior
- Token count is not chained to new chatStore anymore, instead its calculated with groupProjects. (if needed can use projectStore.getTotalTokens)
image

- Due to (2), thus this need to take care when "click task to replay":
image

- Home page displays grouped tasks:
image

- HistorySidebar displays grouped tasks:
image

### Todos:
- Merge #644 (update the server too)
- TaskId == ProjId for old tasks, so now when running new sync_step we are saving to task_id. Sometimes the last task might end in error. System should recover from it.
- SearchHistoryDialog.tsx
```
const handleDelete = (taskId: string) => {
  // TODO: Implement delete functionality similar to HistorySidebar
  console.log("Delete task:", taskId);
};

const handleShare = (taskId: string) => {
  // TODO: Implement share functionality similar to HistorySidebar
  console.log("Share task:", taskId);
};
```
- ⚠️ Further test with different cases of replay (success, error etc..)

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


a7m-1st:
Does anyone also face this issue? Its quite cumbersome

image

src/components/GroupedHistoryView/index.tsx
```diff
@@ -0,0 +1,138 @@
+import React, { useState, useEffect } from "react";
+import { motion, AnimatePresence } from "framer-motion";
+import { ProjectGroup as ProjectGroupType } from "@/types/history";
+import { fetchGroupedHistoryTasks } from "@/service/historyApi";
+import ProjectGroup from "./ProjectGroup";
+import { useTranslation } from "react-i18next";
+import { Loader2, FolderOpen } from "lucide-react";
+import { update } from "electron/main/update";
```
Wendong-Fan: `update` is not used

src/components/GroupedHistoryView/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const onDelete = (historyId: string ) => {
+    try {
+      onTaskDelete(historyId, () => {
+        setProjects(prevProjects => {
+          return prevProjects.map(project => {
+            project.tasks = project.tasks.filter(task => String(task.id) !== historyId);
+            return project;
+          }).filter(project => project.tasks.length > 0);
+        });
+      });
```
Wendong-Fan, line 56: would be more safe to create new project objects using spread operator

src/service/historyApi.ts
```diff
@@ -0,0 +1,117 @@
...
+export const fetchGroupedHistoryTasks = async (setProjects: React.Dispatch<React.SetStateAction<ProjectGroup[]>>) => {
+  try {
+    const res = await proxyFetchGet(`/api/chat/histories/grouped?include_tasks=true`);
+    if(res.status !== 200) {
+      fetchGroupedHistoryTasksLegacy(setProjects);
+    } else setProjects(res.projects);
+  } catch (error) {
+    console.error("Failed to fetch grouped history summaries:", error);
+    setProjects([]);
+  }
```
Wendong-Fan: `proxyFetchGet` returns the parsed JSON body, not the `Response` object, checking `res.status !== 200` seems problematic

src/service/historyApi.ts
```diff
@@ -0,0 +1,117 @@
...
+};
+
+// Function to fetch grouped history summaries only (without individual tasks for better performance)
+export const fetchGroupedHistorySummaries = async (setProjects: React.Dispatch<React.SetStateAction<ProjectGroup[]>>) => {
+  try {
+    const res = await proxyFetchGet(`/api/chat/histories/grouped?include_tasks=false`);
+    if(res.status !== 200) {
```
Wendong-Fan: same here

Wendong-Fan (commented):
thanks @a7m-1st , added enhance PR: https://github.com/eigent-ai/eigent/pull/650

src/components/GroupedHistoryView/index.tsx
a7m-1st: I overlooked it u r right

src/service/historyApi.ts
a7m-1st: Didn't realize this, thanks

PR #661, opened 2025-11-15
fixed image preview by replacing localfile:// with dataURL loader and…

schrodingers-cat-T10:
# Description

This PR fixes the issue where generated chart images were not displayed inside the app and instead appeared as blocked localfile:// URLs, causing the Electron renderer to download the image instead of showing it.

# What this PR does

Adds a new IPC handler: read-file-dataurl, which reads image files from disk and returns them as a Base64 data URL.

Updates the preload script to expose electronAPI.readFileAsDataUrl.

Replaces the old localfile://-based  logic with a new ImageLoader component that uses the data URL provided by Electron.

Ensures all image types (png, jpg, svg, etc.) render correctly in the File Viewer.

# Root Cause

Electron's renderer does not allow loading raw filesystem paths with localfile://, so images returned from generated tasks were blocked with:

``` bash
GET localfile://... net::ERR_UNKNOWN_URL_SCHEME 
```
# screenshot
Screenshot 2025-11-15 at 12 57 06 PM

- [x] Bug fix  
- [ ] New Feature  
- [ ] Documentation update  
- [ ] Other

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -16,6 +16,7 @@ import { copyBrowserData } from './copy'
 import { findAvailablePort } from './init'
 import kill from 'tree-kill';
 import { zipFolder } from './utils/log'
+import mime from "mime";
 import axios from 'axios';
 import FormData from 'form-data';
 import { checkAndInstallDepsOnUpdate, PromiseReturnType, getInstallationStatus } from './install-deps'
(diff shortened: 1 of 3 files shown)


schrodingers-cat-T10:
@Wendong-Fan  and @FooFindBar  , could you guys review it please

PR #664, opened 2025-11-16
Fix flicker sidebox

schrodingers-cat-T10:
### Description

This PR fixes two issues related to image rendering inside the file preview panel:

2. **The preview dropdown flickered horizontally on first open**, because it rendered before its anchor position was calculated, causing a visible jump from `left: 0` to the final position.

#### ✔ Fixed dropdown flicker
- Prevented rendering the dropdown until `anchorStyle` is available.
- Avoids the initial "jump" caused by rendering at `left: 0` before the real position is known.

- [x] Bug fix  
- [ ] New Feature  
- [ ] Documentation update  
- [ ] Other


schrodingers-cat-T10:
@Douglasymlai @Wendong-Fan , hello could you please review this pr

Douglasymlai:
Hi @schrodingers-cat-T10, thanks for the PR. I ran a quick test on your branch and noticed that the flashing issue still occurs. I believe additional updates to HistorySidebar.tsx may be needed.

- `useLayoutEffect` runs synchronously before paint, so the position is set before the first render.
- Conditional rendering with `anchorStyle` ensures the dropdown only renders when positioned.
- `initial={false}` prevents Framer Motion from animating from a default state.
- Resetting on close ensures a fresh calculation each time.


### Suggested changes

#### 1. Use `useLayoutEffect` instead of `useEffect`

Calculate the anchor position synchronously before paint to avoid the flash.

**Current:**
```tsx
useEffect(() => {
  // ... anchor calculation
}, [isOpen]);
```

**Should be:**
```tsx
useLayoutEffect(() => {
  // ... anchor calculation
}, [isOpen]);
```

Also update the import:
```tsx
import { useEffect, useLayoutEffect, useRef, useState } from "react";
```

#### 2. Ensure conditional rendering includes `anchorStyle` check

**Current:**
```tsx
{isOpen && (
```

**Should be:**
```tsx
{isOpen && anchorStyle && (
```

#### 3. Remove fallback values from style prop

Since we only render when `anchorStyle` exists, remove the fallbacks.

**Current:**
```tsx
style={{
  left: anchorStyle ? anchorStyle.left : 0,
  top: anchorStyle ? anchorStyle.top : 40,
}}
```

**Should be:**
```tsx
style={{
  left: anchorStyle.left,
  top: anchorStyle.top,
}}
```

#### 4. Disable initial animation on the dropdown `motion.div`

Set `initial={false}` to prevent Framer Motion from animating from an initial state.

**Current:**
```tsx
<motion.div
  initial={{ y: -8, opacity: 0 }}
  animate={{ y: 0, opacity: 1 }}
  // ...
```

**Should be:**
```tsx
<motion.div
  initial={false}
  animate={{ y: 0, opacity: 1 }}
  // ...
```

#### 5. Reset `anchorStyle` when closing

Reset `anchorStyle` to `null` when `isOpen` becomes `false` to ensure a fresh calculation on next open.

**In the `useLayoutEffect`:**
```tsx
if (isOpen) {
  updateAnchor();
  window.addEventListener("resize", updateAnchor);
} else {
  // Reset anchorStyle when closing to prevent stale values
  setAnchorStyle(null);
}
```


---

schrodingers-cat-T10:
alright.. on it , thanks ! i felt like it was fixed when i tried a couple of time . i did the few of the suggested changes earlier  but i guess it didnt fixed , let me add the few other too . thanks for the suggestion @Douglasymlai

schrodingers-cat-T10:
@Douglasymlai hello there , i did the changes you mentioned and tested it out . i feel like its smooth now , can i give the pr with the same branch or should i take a new branch

Douglasymlai:
Thanks, @schrodingers-cat-T10. I think the issue only occurs when opening the app initially. After the position is rendered, it no longer appears.

Yes, you can go ahead and commit under this PR — that should be easier. After that, I’ll run a quick test. @Wendong-Fan can help with the final check.

schrodingers-cat-T10:
@Douglasymlai could test it again ?

schrodingers-cat-T10:
alright.. thank you ! , excited to contribute more ! @Douglasymlai

PR #696, opened 2025-11-18
feat: browser_external_cdp

nitpicker55555:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -157,16 +157,22 @@ async function isCdpPortAlive(port: number): Promise<boolean> {
 /** Run one health-check cycle: remove dead browsers, persist & notify if changed. */
 async function runPoolHealthCheck(): Promise<void> {
   if (cdp_browser_pool.length === 0) return;
+  // Probe a snapshot so add/remove IPC handlers can run safely in parallel.
+  const snapshot = [...cdp_browser_pool];
   const results = await Promise.all(
-    cdp_browser_pool.map((b) => isCdpPortAlive(b.port))
+    snapshot.map((b) => isCdpPortAlive(b.port))
   );
-  const deadPorts: number[] = [];
-  for (let i = results.length - 1; i >= 0; i--) {
-    if (!results[i]) {
-      deadPorts.push(cdp_browser_pool[i].port);
-      cdp_browser_pool.splice(i, 1);
-    }
-  }
+  const deadIds = snapshot
+    .filter((_, idx) => !results[idx])
+    .map((browser) => browser.id);
+  if (deadIds.length === 0) return;
+
+  const deadIdSet = new Set(deadIds);
+  const removedBrowsers = cdp_browser_pool.filter((b) => deadIdSet.has(b.id));
+  if (removedBrowsers.length === 0) return;
+
+  cdp_browser_pool = cdp_browser_pool.filter((b) => !deadIdSet.has(b.id));
+  const deadPorts = removedBrowsers.map((b) => b.port);
   if (deadPorts.length > 0) {
     log.info(
       `[CDP POOL] Health-check removed dead ports: ${deadPorts.join(', ')}. pool_size=${cdp_browser_pool.length}`
@@ -178,6 +184,10 @@ async function runPoolHealthCheck(): Promise<void> {
 
 /** Start periodic health check (call after window is created). */
 function startCdpHealthCheck(): void {
+  if (cdpHealthCheckTimer) {
+    clearInterval(cdpHealthCheckTimer);
+    cdpHealthCheckTimer = null;
+  }
   log.info('[CDP POOL] Starting health check (interval=3s)');
   // Run once immediately
   runPoolHealthCheck();
@@ -3429,6 +3439,9 @@ app.whenReady().then(async () => {
 // ==================== window close event ====================
 app.on('window-all-closed', () => {
   log.info('window-all-closed');
+
+  // Stop polling when no window is open (important on macOS reopen flow).
+  stopCdpHealthCheck();
 
   // Clean up WebView manager
   if (webViewManager) {
--- a/src/pages/Dashboard/Browser.tsx
+++ b/src/pages/Dashboard/Browser.tsx
@@ -233,11 +233,20 @@ export default function Browser() {
 
       // Port is alive — add to CDP pool
       if (window.electronAPI?.addCdpBrowser) {
-        await window.electronAPI.addCdpBrowser(
+        const addResult = await window.electronAPI.addCdpBrowser(
           portNum,
           true,
           `External Browser (${portNum})`
         );
+        if (!addResult?.success) {
+          setConnectError(
+            addResult?.error || t('layout.failed-to-add-browser')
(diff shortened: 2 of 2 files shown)


nitpicker55555:
Need this camel pr https://github.com/camel-ai/camel/pull/3666/files

src/store/chatStore.ts
```diff
@@ -605,6 +623,8 @@ const chatStore = (initial?: Partial<ChatStore>) => createStore<ChatStore>()(
 					summary_prompt: ``,
 					new_agents: [...addWorkers],
 					browser_port: browser_port,
+					use_external_cdp: use_external_cdp,
+					cdp_browsers: cdp_browsers,
 					env_path: envPath,
```
a7m-1st: @nitpicker55555 I can see that use_external_cdp is passed as a boolean to the backend for the agents to use, what happens for multi turn tasks?

```
POST /chat/{project_id}
question: "use the browser agent to see what is in my screen?"
task_id : "1768494591185-9818"

File: backend/app/utils/agent.py [hunk 2]
```diff
@@ -71,6 +72,74 @@
 
 NOW_STR = datetime.datetime.now().strftime("%Y-%m-%d %H:00:00")
 
+# Global counter for round-robin browser selection from pool
+_browser_selection_counter = 0
+
```
a7m-1st: Not sure if this is being used in the code?

backend/app/utils/agent.py
```diff
@@ -851,30 +1063,199 @@ def search_agent(options: Chat):
...
+        new_toolkit = HybridBrowserToolkit(
+            options.project_id,
+            headless=False,
+            browser_log_to_file=True,
+            stealth=True,
+            session_id=session_id,
+            log_dir=str(task_log_dir),
+            default_start_url=default_url,
+            connect_over_cdp=True,
```
a7m-1st: Maybe we can use the below?

```suggestion
            connect_over_cdp=options.use_external_cdp,
```

a7m-1st: added in https://github.com/eigent-ai/eigent/pull/849

backend/app/utils/agent.py
```diff
@@ -851,30 +1063,199 @@ def search_agent(options: Chat):
...
             "browser_back",
             "browser_forward",
             "browser_switch_tab",
             "browser_enter",
             "browser_visit_page",
-            "browser_scroll",
-            # "browser_get_som_screenshot",
+            "browser_get_page_snapshot"
+            # "browser_scroll",
+            # "browser_get_page_snapshot",
```
a7m-1st:
```suggestion
            # "browser_scroll",
```

backend/app/utils/workforce.py
```diff
@@ -487,9 +487,115 @@ def stop_gracefully(self) -> None:
...
+                for i, child in enumerate(self._children):
+                    logger.info(f"[WF-CLEANUP-DEBUG] _children[{i}]: type={type(child).__name__}, "
+                               f"has_worker_agent={hasattr(child, 'worker_agent')}, "
+                               f"has_agent_pool={hasattr(child, 'agent_pool')}")
+
+        logger.info(f"[WF-CLEANUP-DEBUG] hasattr(self, 'coordinator_agent'): {hasattr(self, 'coordinator_agent')}")
+        if hasattr(self, 'coordinator_agent'):
+            logger.info(f"[WF-CLEANUP-DEBUG] self.coordinator_agent is None: {self.coordinator_agent is None}")
+            if self.coordinator_agent is not None:
+                logger.info(f"[WF-CLEANUP-DEBUG] coordinator has _cleanup_callback: {hasattr(self.coordinator_agent, '_cleanup_callback')}")
```
a7m-1st: maybe we can remove this?

a7m-1st (commented):
Thanks @nitpicker55555 for the PR, I have added some enhancements in https://github.com/eigent-ai/eigent/pull/849, I figured `ListenerAgent.clone` is the main focus (btw according to the callstack its mostly triggered when `workforce.reset()` is called  :) ) but generally can't find logic issues. 

I analyzed if the cdp clone operation was only happening for search agents only as it should be and bcz of the `_cdp_acquire_callback` it seems search_agent is the only one that has it 👍

backend/app/utils/agent.py
a7m-1st: I think it was enabled by default as to why it was working. My concern is what constitutes enabling the `use_external_cdp`? In this example even though I set a port it doesn't seems to send True.

image

a7m-1st:
# Clone Operation Overview

```mermaid
%%{init: {'theme': 'base', 'themeVariables': { 'primaryColor':'#ffffff', 'primaryBorderColor':'#cccccc', 'background':'#ffffff', 'mainBkg':'#ffffff', 'secondBkg':'#ffffff', 'tertiaryColor':'#ffffff'}}}%%
sequenceDiagram
  participant P as Parent Agent (ListenChatAgent)
  participant M as CDP Manager (_cdp_pool_manager)
  participant TK as Toolkit (HybridBrowserToolkit)
  participant A as New Agent (Cloned)

  Note over P: clone(with_memory=True/False)
  P->>P: Generate Clone ID & unique Session ID
  
  rect rgb(230, 240, 255)
    Note right of P: CDP Pre-acquisition (If callback exists)
    P->>M: acquire_browser(cdp_browsers, new_session_id)
    M-->>P: Return Port (e.g., 9223)
    P->>TK: Temporarily change CDP URL to localhost:9223
  end

  P->>P: _clone_tools() (Deep copy of tool definitions)
  
  rect rgb(255, 240, 230)
    Note right of P: Restore Parent State
    P->>TK: Restore original CDP URL (e.g., localhost:9222)
  end

  P->>A: Instantiate ListenChatAgent with cloned tools
  P->>A: Copy Metadata (task_id, callbacks, options)

  Note over A: Post-Init Configuration
  A->>A: find HybridBrowserToolkit in cloned tools
  A->>A: Set _browser_toolkit & _cleanup_callback

  alt with_memory is True
    P->>P: memory.retrieve() (Get all records)
    P->>A: memory.write_record() (Populate history)
  end

  P-->>User: Return New Agent

backend/app/utils/workforce.py
a7m-1st: removed in https://github.com/eigent-ai/eigent/pull/696

src/store/chatStore.ts
nitpicker55555: use_external_cdp configuration is preserved in multi turn because construct_workforce(options) reuses the same options object, so the use_external_cdp and cdp_browsers settings remain unchanged. But, because the multi-turn workforce is destroyed and reconstructed, the agents will be reassigned to a new browser port from pool.

a7m-1st: ya exactly, after second thoughts if a user closed his browser session after the first run (unless he restarted the kept it open) then technically available ports are dynamically updated.

I guess the small edge case would be if a user configured the cdp after the first turn 🤔

PR #713, opened 2025-11-20
fix: stop task not working as expected

Wendong-Fan:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -89,6 +89,21 @@ app.commandLine.appendSwitch('max_old_space_size', '4096');
 app.commandLine.appendSwitch('enable-features', 'MemoryPressureReduction');
 app.commandLine.appendSwitch('renderer-process-limit', '8');
 
+// ==================== protocol privileges ====================
+// Register custom protocol privileges before app ready
+protocol.registerSchemesAsPrivileged([
+  {
+    scheme: 'localfile',
+    privileges: {
+      standard: true,
+      secure: true,
+      supportFetchAPI: true,
+      corsEnabled: false,
+      bypassCSP: false,
+    },
+  },
+]);
+
 // ==================== app config ====================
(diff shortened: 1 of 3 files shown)


src/store/chatStore.ts
```diff
@@ -1598,13 +1666,29 @@ const chatStore = (initial?: Partial<ChatStore>) => createStore<ChatStore>()(
...
+					// Clean up AbortController on error
+					try {
+						if (activeSSEControllers[newTaskId]) {
+							delete activeSSEControllers[newTaskId];
+						}
+					} catch (error) {
+						console.warn('Error cleaning up AbortController on SSE error:', error);
+					}
 					throw err;
 				},
```
a7m-1st, line 1751: Are we targetting for skip to behave like stop? @Wendong-Fan 
🤔 If so I think we can re-implement the new on "error" message logic on #663.

a7m-1st: The PR is almost done @Wendong-Fan

src/components/ChatBox/index.tsx
```diff
@@ -416,28 +421,38 @@ export default function ChatBox(): JSX.Element {
...
 
-			// Update task status to finished
-			chatStore.setStatus(taskId, 'finished');
+			// Only stop local task if backend call succeeds
+			chatStore.stopTask(taskId);
 			chatStore.setIsPending(taskId, false);
-			
-			// toast.success("Task skipped successfully", {
-			// 	closeButton: true,
-			// });
```
a7m-1st, line 432: @Wendong-Fan in this PR https://github.com/eigent-ai/eigent/pull/735, it just makes the skip task == stop task. 
- Replay proof, where the skip task can be replicated with the replay api.
- But maybe can borrow your idea of activeSSEControllers as it still suffers from ghost SSEs (i.e. Post stop)

PR #743, opened 2025-11-20
enhance: update history fetch state + token count on simple tasks

### LINKED ISSUE #730: [BUG] Running grouped tasks not updating state as opposed to completed ones

0.0.72

### System information

win 11, py 11

### Problem description

the history sidebar while task is running
1) it doesn't reflect the latest title (just new project)
2) The Pin number is 2 instead of 1 (Though those issues are fixed when task is done running)

Image

## After task completed all is good (pin == # of tasks):

Image

### log

na

### Additional context

_No response_

a7m-1st:
### Description
1. Return token updates
2. Use the data from the history api as much as possible (refetches on each update rather than local optimistic updates)

### Known bugs
- When running a new task the "New Project" is actually being replaced. Thus the view on "2" tasks, but it disappears once the title is generated.

https://github.com/user-attachments/assets/602177d3-2fa5-49e7-ab3b-afb69d9224ac

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


LuoPengcheng12138:
Hi, @a7m-1st ,During testing, I noticed that the number of tasks remains 2 before the completion of simple tasks and the splitting of complex tasks.

a7m-1st:
Yes @LuoPengcheng12138, I believe I added it in the video. The reason being is the synchronizing between database & local state.
Where a chat in new state would be counted with the default "New Project" chatStore.

Won't cause any issues, and would be fixed on 'refresh'. But if got the time, definitely would need help on polishing this.

LuoPengcheng12138:
Cool.

Douglasymlai:
This issue has been resolved in another PR. Will close this PR for now.

PR #765, opened 2025-11-24
test: add feature testing guideline and example

Wendong-Fan:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/test/feature/SendFirstMessage.feature.test.tsx
+++ b/test/feature/SendFirstMessage.feature.test.tsx
@@ -1,0 +1,239 @@
+/**
+ * Feature Testing Sample Documentation
+ *
+ * ============================================================================
+ * What is Integration/Feature Testing?
+ * ============================================================================
+ *
+ * Feature tests focus on what users can see and do, rather than how the code is implemented.
+ *
+ * ## Feature Tests vs Unit Tests
+ *
+ * ### Unit Tests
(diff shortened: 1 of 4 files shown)


test/feature/SendFirstMessage.feature.test.tsx
```diff
@@ -182,6 +181,109 @@ describe('Feature test example: chat experience', () => {
...
+      expect(screen.getByText('Todo List Application')).toBeInTheDocument()
+      expect(screen.getByText('Create HTML structure for todo list')).toBeInTheDocument()
+      expect(screen.getByText('Implement JavaScript functionality')).toBeInTheDocument()
+      expect(screen.getByText('Add CSS styling')).toBeInTheDocument()
+    })
+
+    // 7. Verify chatStore state is correct by getting fresh state
+    const updatedChatStore = result.current.chatStore
+    expect(updatedChatStore!.tasks[taskId!].summaryTask).toContain('Todo List Application')
+    expect(updatedChatStore!.tasks[taskId!].taskInfo).toHaveLength(3)
```
Wendong-Fan, line 285: let's  assert user-facing behavior, not implementation details, part 7 is actually checking internal state, which is not necessary, we'd better add mock in SSE level to do feature test

PR #772, opened 2025-11-27
chore：fix windows uv 

fengju0213:
Problem: Backend startup fails on Windows after a system restart with the error Backend environment check failed: No Python at .... This occurs because uv cannot locate the Python environment specified during installation. Additionally, the existing auto-repair mechanism fails to recover when the Python installation is corrupted (i.e., the directory exists but python.exe is missing).

Changes:

Consistent Environment Variables: Explicitly injected UV_PYTHON_INSTALL_DIR and UV_TOOL_DIR in 
startBackend
 to ensure the runtime configuration matches the installation configuration, allowing uv to correctly locate the cached Python environment.
Enhanced Auto-Repair: Improved the repair logic upon pre-flight check failure:
Added detection and cleanup for corrupted Python installations (where the directory exists but the executable is missing).
Added an explicit uv python install step before running uv sync to ensure the Python interpreter is correctly reinstalled.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/init.ts
+++ b/electron/main/init.ts
@@ -21,16 +21,16 @@ export function getMainWindow(): BrowserWindow | null {
 export async function checkToolInstalled() {
     return new Promise<PromiseReturnType>(async (resolve, reject) => {
         if (!(await isBinaryExists('uv'))) {
-            resolve({success: false, message: "uv doesn't exist"})
+            resolve({ success: false, message: "uv doesn't exist" })
             return
         }
 
         if (!(await isBinaryExists('bun'))) {
-            resolve({success: false, message: "Bun doesn't exist"})
+            resolve({ success: false, message: "Bun doesn't exist" })
             return
         }
 
-        resolve({success: true, message: "Tools exist already"})
+        resolve({ success: true, message: "Tools exist already" })
     })
 
 }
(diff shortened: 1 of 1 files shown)


electron/main/init.ts
```diff
@@ -202,10 +204,47 @@ export async function startBackend(setPort?: (port: number) => void): Promise<an
...
-        } catch (testErr) {
-            log.error(`Pre-flight check failed: ${testErr}`);
-            reject(new Error(`Backend environment check failed: ${testErr}`));
-            return;
+        } catch (testErr: any) {
+            log.warn(`Pre-flight check failed, attempting repair: ${testErr}`);
+
+            try {
+                // Check if python executable exists but is corrupted
+                const pythonPath = path.join(getCachePath('uv_python'), 'cpython-3.10.16-windows-x86_64-none', 'python.exe');
```
Wendong-Fan, line 212: this path only works on Windows x86_64. The repair logic will not work correctly on macos

electron/main/init.ts
```diff
@@ -165,6 +165,8 @@ export async function startBackend(setPort?: (port: number) => void): Promise<an
         PYTHONIOENCODING: 'utf-8',
         PYTHONUNBUFFERED: '1',
         UV_PROJECT_ENVIRONMENT: venvPath,
+        UV_PYTHON_INSTALL_DIR: getCachePath('uv_python'),
+        UV_TOOL_DIR: getCachePath('uv_tool'),
```
Wendong-Fan, line 169: The UV environment variables are now defined separately in both `install-deps.ts` and `init.ts`, which could lead to inconsistencies if one is updated but not the other

PR #777, opened 2025-12-01
update gemini docs

Douglasymlai:
Added Gemini Blog

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


backend/app/controller/electron_browser.cjs
fengju0213, line 1: I don't think this file is needed when modifying the document?

PR #812, opened 2025-12-30
enhance: tool info stream UI

Wendong-Fan:
Description:

  Previously, during task splitting, the streaming output displayed raw text with <task> XML tags, which was inconsistent with
  the final TaskCard UI style.

  Changes:
  - Created StreamingTaskList component that parses <task> tags and displays tasks in a list format
  - Matches TaskCard styling (background, rounded corners, progress bar, task indicators)
  - Shows loading spinner for tasks currently being streamed
  - Added i18n support for labels

  Before: Raw XML tags displayed in a plain text block
  After: Parsed tasks displayed in a styled card matching the final TaskCard appearance

### DIFF
--- a/src/components/ChatBox/TaskBox/StreamingTaskList.tsx
+++ b/src/components/ChatBox/TaskBox/StreamingTaskList.tsx
@@ -1,0 +1,120 @@
+import { useMemo } from "react";
+import { CircleDashed, LoaderCircle } from "lucide-react";
+import { useTranslation } from "react-i18next";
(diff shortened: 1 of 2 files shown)


fengju0213:
I think the previous visual effects might have been better?

Wendong-Fan:
> I think the previous visual effects might have been better?

hey @fengju0213 , I just discussed with @Douglasymlai , we agreed to merge this version first, later @Douglasymlai would do further enhancement for the UI

PR #823, opened 2026-01-12
feat: schedule and webhook triggers

### LINKED ISSUE #1284: [Feature Request] Test and Go-live with Trigger feature
### Motivation

We are launching the Trigger feature, which allows users to configure schedule triggers and event triggers for each project. We are now inviting the community to participate in branch testing.

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

a7m-1st:
### Description
- Depends on the sever side implementation
- Introduced the Trigger feature
image

### Highlights and upgrades
UX - Click twice to edit trigger (in addition to edit button)
UX - Added a webhook dialog after its creation
UX - Live notifications shows all the logs from all triggers (can be modified)
Logic - A single websocket (if trigger is available) per session for all projects
Server - Execution Event marked as missed if not acknowledged in 60 seconds

### To dos
- [x] UI
- [x] Integrate CRUD into UI
- [x] Integrate Real time changes
- [x] Run the task in foreground project
- [x] Add concurrent tasks to queue
- [x] Run each trigger and project in the background (need to update handleSend logic)

### Small issues
- [x] When clicking to Edit trigger it is not consistent (fields empty)
- [x] All time sent to client is utc based - need to convert to local time in client
- [x] Custom crons doesn't update next run at in UI.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


src/store/triggerTaskStore.ts
```diff
@@ -0,0 +1,265 @@
...
+ * Formats the task message with context from webhook or schedule
+ */
+export function formatTriggeredTaskMessage(task: TriggeredTask): string {
+    const parts: string[] = [];
+    
+    // Add the main task prompt
+    parts.push(task.taskPrompt);
+    
+    // Add context separator if we have input data
+    if (task.inputData && Object.keys(task.inputData).length > 0) {
```
4pmtong: We're directly using `inputData` without sanitization. This might cause`Input injection risks`, can try sanitizeInputData with `body size limit` and use `DOMPurify` for sanitization, also can try to filter some sensitive fields like password, token etc.

src/store/triggerStore.ts
```diff
@@ -0,0 +1,128 @@
...
+    triggerReconnect: () => {
+        const callback = get().wsReconnectCallback;
+        if (callback) {
+            callback();
+        }
+    },
+
+    addTrigger: (triggerData: Partial<Trigger>) => {
+        const newTrigger: Trigger = {
+            id: triggerData.id || Date.now(),
```
4pmtong: how about use `Date.now() + a random key`?

src/service/triggerApi.ts
```diff
@@ -0,0 +1,288 @@
...
+  try {
+    const res = await proxyFetchGet(`/api/trigger/${triggerType}/config`);
+    return res;
+  } catch (error) {
+    console.error("Failed to fetch trigger config:", error);
+    throw error;
+  }
+};
+
+export const proxyCreateTrigger = async (triggerData: TriggerInput): Promise<Trigger> => {
```
4pmtong: how about add size limit here?

src/store/triggerStore.ts
a7m-1st: Nice catch, just removed it, id is never supposed so be null anyways

src/service/triggerApi.ts
a7m-1st: Would have been preferable to use schema validation with Zod for these cases I think, but for name, description and task_prompt I added validation in the server side too. For the dynamic fields we can add validation via:
```
"ui:validation"?: "regex" | "email" | "url" | "number";
    minLength?: number;
    maxLength?: number;
    pattern?: string;
```

For now just added maxLength to the Input fields

image

src/store/triggerTaskStore.ts
a7m-1st: You are right, the only susceptible triggers are the webhook and app trigger. Those inputData are recieved via websockets (live events received by the server). 

For the event body size and filtering, yes that we should block in the serverside I think, as I don't think the  would be good to propagate TriggerExecution event to the client in that sense.

As for the dom sanitization, it is only made for DOM scripts though, and I don't think we will render the input as part of DOM.

4pmtong:
hey @a7m-1st The server changes seem to be mostly license headers, could you help check it?

4pmtong (approved):
Approve for testing

PR #873, opened 2026-01-16
Disable parallel tool calls for search agent & fix azure api

LuoPengcheng12138:
### Description

Only `camel/configs/openai_config.py` and `camel/configs/aihubmix_config.py` declare `parallel_tool_calls`. Therefore, the only explicitly supported platforms are OpenAI/Azure OpenAI (which share the same configuration) and AihubMix (corresponding to `camel/models/openai_model.py`, `camel/models/azure_openai_model.py` and `camel/models/aihubmix_model.py`).
### What is the purpose of this pull request? 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

PR #849, opened 2026-01-14
enhance: browser external cdp a7m

a7m-1st:
### Description
Hi there @nitpicker55555 , went through a quick review of the electron side and it looks good, I just need some time to check the backend logic. I will be committing to this branch.

### What is the purpose of this pull request? 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

PR #879, opened 2026-01-16
:bug: fix the hybrid browser toolkit issue

4pmtong:
#### 1. Fix Azure OpenAI Model Initialization

Fixed two issues when using Azure OpenAI models:

- **Missing `api_version`**: Now properly passed to `ModelFactory.create()` as initialization parameter
- **Incorrect `azure_deployment_name` usage**: Fixed parameter being incorrectly passed to API calls instead of model initialization

**Solution**: Updated `agent_model()` to separate Azure initialization parameters (`api_version`, `azure_deployment_name`) from API call parameters. Azure params are now passed as direct kwargs to `ModelFactory.create()`.

#### 2. Update HybridBrowserToolkit Connection Mode

Changed `HybridBrowserToolkit` initialization in `search_agent()` to use `connect_over_cdp=False` instead of CDP connection mode.

**Related Files**:
- `backend/app/utils/agent.py`:
  - Updated `agent_model()` to handle Azure OpenAI parameters correctly
  - Modified `search_agent()` to set `connect_over_cdp=False` for `HybridBrowserToolkit`

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/utils/agent.py
+++ b/backend/app/utils/agent.py
@@ -8,11 +8,20 @@ from typing import Any, Callable, Dict, List, Tuple
 import uuid
 from utils import traceroot_wrapper as traceroot
 from camel.agents import ChatAgent
-from camel.agents.chat_agent import StreamingChatAgentResponse, AsyncStreamingChatAgentResponse
+from camel.agents.chat_agent import (
+    StreamingChatAgentResponse,
+    AsyncStreamingChatAgentResponse,
+)
 from camel.agents._types import ToolCallRequest
 from camel.memories import AgentMemory
 from camel.messages import BaseMessage
-from camel.models import BaseModelBackend, ModelFactory, ModelManager, OpenAIAudioModels, ModelProcessingError
+from camel.models import (
+    BaseModelBackend,
+    ModelFactory,
+    ModelManager,
+    OpenAIAudioModels,
+    ModelProcessingError,
+)
 from camel.responses import ChatAgentResponse
 from camel.terminators import ResponseTerminator
 from camel.toolkits import FunctionTool, RegisteredAgentToolkit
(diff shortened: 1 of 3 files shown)


nitpicker55555:
Hi，can you successfully run browsertoolkit after modification？

backend/app/utils/agent.py
```diff
@@ -858,7 +957,8 @@ def search_agent(options: Chat):
         stealth=True,
         session_id=str(uuid.uuid4())[:8],
         default_start_url="about:blank",
-        cdp_url=f"http://localhost:{env('browser_port', '9222')}",
+        # cdp_url=f"http://localhost:{env('browser_port', '9222')}",
+        connect_over_cdp=False,
         enabled_tools=[
```
a7m-1st, line 962: Thanks @4pmtong , just to mention though according to cdp branch and #849 connect_over_cdp is updated according to the user configuration

4pmtong: "Thanks @a7m-1st ! Good catch on the `connect_over_cdp` setting. then how about the Azure configuration?

4pmtong:
> Hi，can you successfully run browsertoolkit after modification？
Yes, it's running successfully.

backend/app/utils/agent.py
a7m-1st: I think we can still merge it though until the cdp branch is merged : )

Regarding azure I have reviewed a similar PR, does this fix it?
https://github.com/eigent-ai/eigent/pull/873

a7m-1st (approved):
All in all if ever other thing is just refactoring then we can have a patch for the hybrid browser till the feature branch is merged to sync with latest camel version.

backend/app/utils/agent.py
4pmtong: Sounds good @a7m-1st! Let's merge this PR first. I'll review #873 for the Azure fix and we can coordinate the integration. Thanks for flagging these! 👍

nitpicker55555:
Close since we better resubmit an updated PR for the linter changes after the other features have been merged.

PR #892, opened 2026-01-16
Add system diagnostics health endpoint and UI

RickyVishwakarma:
## What does this PR do?

Adds a system diagnostics feature to make it easier to check if the backend
is ready before running workflows.

### Backend
- Extends the existing `/health` endpoint to report:
  - Database connectivity
  - Model configuration
  - Tool registry availability

### Frontend
- Adds a `/diagnostics` page to show system status
- Protects the page behind authentication
- Uses existing routing and layout

---

## Why is this needed?

Right now there is no easy way to verify whether the system
is fully set up. This helps with:

- Debugging setup issues
- Onboarding new users
- Understanding runtime state

---

## How was this tested?

- Checked `/health` endpoint response
- Loaded diagnostics page when logged in
- Verified redirect to login when logged out


RickyVishwakarma:
Hi maintainers 👋  
This PR is ready for review.  
Could someone please approve the workflow and review the changes when convenient?  
Thanks!

Wendong-Fan:
thanks @RickyVishwakarma for the contribution!

backend/app/controller/health_controller.py
```diff
@@ -1,22 +1,104 @@
...
-    response = HealthResponse(status="ok", service="eigent")
-    logger.debug("Health check completed", extra={"status": response.status, "service": response.service})
-    return response
 
+    # -------------------------------
+    # Database check
+    # -------------------------------
+    db_status = "unknown"
+    try:
+        from app.database import engine  # adjust path ONLY if needed
```
Wendong-Fan: seems in app no database

backend/app/controller/health_controller.py
```diff
@@ -1,22 +1,104 @@
...
+    model_name = os.getenv("MODEL_NAME")
+
+    model_configured = bool(model_api_key and model_name)
+
+    # -------------------------------
+    # Tool availability check
+    # -------------------------------
+    tools_available = False
+    try:
+        from app.tools.registry import TOOL_REGISTRY  # adjust path if needed
```
Wendong-Fan: no TOOL_REGISTRY in our project

backend/app/controller/health_controller.py
```diff
@@ -1,22 +1,104 @@
...
+        db_status = "ok"
+    except Exception as e:
+        logger.warning("Database health check failed", extra={"error": str(e)})
+        db_status = "error"
+
+    # -------------------------------
+    # Model configuration check
+    # -------------------------------
+    model_api_key = os.getenv("MODEL_API_KEY")
+    model_name = os.getenv("MODEL_NAME")
```
Wendong-Fan: no `MODEL_API_KEY`, no `MODEL_NAME `

src/pages/Diagnostics.tsx
```diff
@@ -0,0 +1,57 @@
...
+  model: { configured: boolean };
+  tools: { available: boolean };
+};
+
+export default function Diagnostics() {
+  const [data, setData] = useState<HealthResponse | null>(null);
+  const [error, setError] = useState("");
+
+  useEffect(() => {
+    fetch("/health")
```
Wendong-Fan: i think use ` /health` would not be able to access to the right backend

Wendong-Fan (commented):
thanks for the contribution, left some comments below

backend/app/controller/health_controller.py
RickyVishwakarma: > no TOOL_REGISTRY in our project

Thanks for pointing this out! You’re right — I assumed a tool registry that doesn’t exist in the current codebase. I’ll remove this check and keep the health endpoint aligned with existing project structures.

backend/app/controller/health_controller.py
RickyVishwakarma: > no `MODEL_API_KEY`, no `MODEL_NAME `

Got it, thanks for clarifying. I’ll remove the model environment variable checks and avoid making assumptions about model configuration in this endpoint.

src/pages/Diagnostics.tsx
RickyVishwakarma: > i think use ` /health` would not be able to access to the right backend

Thanks for the note — makes sense. I’ll update the frontend to use the existing backend access pattern instead of calling `/health` directly.

RickyVishwakarma:
Thanks for the review! I’ve updated the diagnostics page to use the proper backend path and removed assumptions about model and tool fields so it aligns with the current health endpoint.

backend/app/controller/health_controller.py
```diff
@@ -1,22 +1,51 @@
...
+    Health check endpoint for verifying backend and database readiness.
+    """
+
     logger.debug("Health check requested")
-    response = HealthResponse(status="ok", service="eigent")
-    logger.debug("Health check completed", extra={"status": response.status, "service": response.service})
-    return response
 
+    # Database connectivity check
+    db_status = "unknown"
```
bytecii: Can we make these status as string enum?

src/pages/Diagnostics.tsx
```diff
@@ -0,0 +1,45 @@
+import { useEffect, useState } from "react";
```
bytecii, line 8: Do we need this as a frontend page here for now? IMO we should have a page in https://www.eigent.ai/ to indicate the health. cc @Wendong-Fan

Wendong-Fan: hey @bytecraftii , thanks for the feedback! The diagnostics page is meant to check local deployment status (database connectivity, model config, etc.) and  https://www.eigent.ai/ can't access users' local environments

Also since we're fully open source and committed to "build in public", let's keeping features like this in the open codebase rather than on a separate closed site

a cloud service status page on https://www.eigent.ai/ could be a separate addition if needed in the future

cc @lightaime

bytecii: I see

RickyVishwakarma: Thanks both for the clarification — that makes sense.

backend/app/controller/health_controller.py
```diff
@@ -14,23 +14,52 @@
...
+    try:
+        from app.database import engine  # existing project database engine
```
Wendong-Fan: This import will always fail because there is no `app.database` module in `backend/`. The `backend/` is the local agent backend and has no database.

The database engine lives in `server/app/component/database.py` (the self-deploy server). If the goal is to add a DB health check for self-deploy users, this should be added to `server/app/controller/health_controller.py` instead, with:
```python
from app.component.database import engine
```
As it stands, this silently catches the `ImportError` and always reports `db_status = "error"`, which is misleading.

backend/app/controller/health_controller.py
```diff
@@ -14,23 +14,52 @@
...
+        with engine.connect() as conn:
```
Wendong-Fan: Minor note: `engine.connect()` is a synchronous call inside an `async def` handler, which would block the event loop. If this gets moved to `server/`, consider either making the handler a regular `def` (FastAPI will run it in a threadpool automatically) or using `run_in_executor`.

src/pages/Diagnostics.tsx
```diff
@@ -0,0 +1,45 @@
...
+  };
+};
+
+export default function Diagnostics() {
+  const [data, setData] = useState<HealthResponse | null>(null);
+  const [error, setError] = useState("");
+
+  useEffect(() => {
+    fetch("/api/health")
+      .then((res) => {
```
Wendong-Fan: The project uses centralized API utilities (`fetchGet` / `proxyFetchGet` from `@/api/http`) for all backend calls. Using raw `fetch("/api/health")` here will:
1. Not resolve the correct backend URL (Electron uses a dynamic port via `getBaseURL()`)
2. Not include auth headers
3. Not work in production

Should be something like:
```ts
import { fetchGet } from "@/api/http";

fetchGet("/health").then(setData).catch(...);
```

src/pages/Diagnostics.tsx
Wendong-Fan, line 8: Missing the Apache 2.0 license header that all source files in this project use. Please add the standard copyright block at the top.

Wendong-Fan (changes requested):
Thanks for working on the diagnostics feature! The idea of surfacing system health info is great. I found a few issues that need to be addressed before this can be merged.

backend/app/controller/health_controller.py
burnsaustin145: To add more detail: the project does not use `MODEL_API_KEY` or `MODEL_NAME` as environment variable names. Model credentials are request-scoped, not globally configured — they are passed per-request through the `Chat` Pydantic model in `app/model/chat.py` as `api_key`, `model_platform`, and `model_type` fields.

Checking env vars with these names would always return empty, reporting the model as unconfigured even for a fully working user. That validation already happens at task submission time (the `Chat` model requires those fields). There is no global model config for a health endpoint to probe here.

RickyVishwakarma:
Thanks for the review! I’ve addressed the comments and updated the backend and diagnostics page accordingly. Also resolved the merge conflicts.

RickyVishwakarma:
Thanks for the detailed review! I’ve updated the health endpoint to align with the backend structure, removed the database-related logic, and switched the diagnostics page to use the existing API utilities. I also cleaned up the routing changes and resolved conflicts. Please let me know if anything else needs adjustment.

PR #920, opened 2026-01-17
Add Linux Support

georgepanther3:
added linux support so you can build on linux now. also added github actions workflow to auto-build releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "build:mac:test": "npm run preinstall-deps && npm run clean-symlinks && npm run compile-babel && tsc && vite build && electron-builder --mac && npm run test-signing",
     "build:win": "npm run preinstall-deps && npm run compile-babel && tsc && vite build && electron-builder --win",
     "build:linux": "npm run preinstall-deps && npm run clean-symlinks && npm run compile-babel && tsc && vite build && electron-builder --linux",
-    "build:all": "npm run preinstall-deps && npm run compile-babel && tsc && vite build && electron-builder --mac --win",
+    "build:all": "npm run preinstall-deps && npm run compile-babel && tsc && vite build && electron-builder --mac --win --linux",
     "preview": "vite preview",
     "pretest": "vite build --mode=test",
     "test": "vitest run",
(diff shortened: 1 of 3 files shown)


brycekt887:
Pretty cool addition.  Was there anything tricky about getting the Linux build working compared to the other popular platforms?

Wendong-Fan:
thanks for the contribution! maybe @bytecraftii and @fengju0213 could also checking this

georgepanther3:
> Pretty cool addition. Was there anything tricky about getting the Linux build working compared to the other popular platforms?

Honestly just the uv download. The Linux tarball is nested weirdly compared to Windows, so I had to tweak the preinstall script to grab the binary. The rest was standard Electron stuff.

fengju0213:
@georgepanther3  thanks for great work！I haven’t run tests yet. One concern: scripts/preinstall-deps.js moves the first uv-* dir it finds under BIN_DIR into uv. If an old uv-* dir is left around, it could overwrite the newly downloaded binary. Consider only moving when uv doesn’t exist, or track the exact extracted directory.Also, Linux AppImage builds on ubuntu-latest often need libfuse2; the workflow doesn’t install it. Might be worth adding a step or confirming the runner already has it. I will test it and then give you further feedback.

package.json
```diff
@@ -23,6 +23,7 @@
     "build:mac": "npm run preinstall-deps && npm run clean-symlinks && npm run compile-babel && tsc && vite build && electron-builder --mac",
     "build:mac:test": "npm run preinstall-deps && npm run clean-symlinks && npm run compile-babel && tsc && vite build && electron-builder --mac && npm run test-signing",
     "build:win": "npm run preinstall-deps && npm run compile-babel && tsc && vite build && electron-builder --win",
+    "build:linux": "npm run preinstall-deps && npm run clean-symlinks && npm run compile-babel && tsc && vite build && electron-builder --linux",
     "build:all": "npm run preinstall-deps && npm run compile-babel && tsc && vite build && electron-builder --mac --win",
```
bytecii: Should we also add --linux here?

nemesifier:
Looking forward to info on how to run this! :+1:

Wendong-Fan:
hey @nemesifier , feel free to download the packaged version from https://github.com/eigent-ai/eigent/actions/runs/21258341352 and take a try, let us know if there's any issue!

Issue #891, opened 2026-01-16
[BUG]


ajmeese7:
### What version of eigent are you using?

0.0.79

### System information

```
OsName              : Microsoft Windows 11 Enterprise
OsVersion           : 10.0.26200
```

### Problem description

The installer fails on Windows machines where the username has a space in it, the file path likely needs to be quoted.

### log

```sh
Error: Backend environment check failed: Error: Command failed: C:\Users\Aaron Meese\.eigent\bin\uv.exe --version 'C:\Users\Aaron' is not recognized as an internal or external command, operable program or batch file. Repair failed: Error: Command failed: C:\Users\Aaron Meese\.eigent\bin\uv.exe python install 3.10 'C:\Users\Aaron' is not recognized as an internal or external command, operable program or batch file.
```

fengju0213:
thanks for reporting this issue ,we fix this issue in https://github.com/eigent-ai/eigent/pull/940 ,free feel to contact us if you have any further questions

PR #940, opened 2026-01-18
[Fix]installer failure on Windows with spaces in path

sashimikun:
### Description

This PR fixes a bug where the installer fails on Windows machines when the user profile path contains spaces (e.g., `C:\Users\John Doe`). The issue was caused by the `uv` executable path not being quoted when executing `npm` and `npx` commands with `shell: true`. Without quotes, the shell interprets the space as an argument separator. This change adds quotes around the `uv` path to ensure it is treated as a single argument. Additionally, it includes a fix for the test mocks to ensure unit tests pass.

Address Issue: https://github.com/eigent-ai/eigent/issues/891

### Verification:

- Created a reproduction script using spawn with shell: true and a path with spaces, confirming that quoting fixes the issue.
- Ran the unit tests (test/unit/electron/install-deps.test.ts) and they all pass.

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Wendong-Fan:
thanks @sashimikun for the contribution! could @4pmtong @LuoPengcheng12138 help checking this?

4pmtong (approved):

PR #949, opened 2026-01-19
chore: Fix docker build

fengju0213:
Cause: During Docker builds, uv bytecode compilation can hit os error 24 on low
  nofile limits, and apt-get occasionally fails to fetch from deb.debian.org with
  502s, breaking the build.
Fix: Set UV_COMPILE_BYTECODE=0 to avoid build-time FD exhaustion, and add retry
options to apt-get update (with --no-install-recommends) to improve build stability.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -4,7 +4,7 @@ FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
 # Install the project into `/app`
 WORKDIR /app
 
-# Disable bytecode compilation to avoid EMFILE during build on low nofile limits
+# Disable bytecode transfer during compilation to avoid EMFILE during build on low nofile limits
 ENV UV_COMPILE_BYTECODE=0
 
 # Copy from the cache instead of linking since it's a mounted volume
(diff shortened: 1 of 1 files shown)


server/Dockerfile
```diff
@@ -4,8 +4,8 @@ FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
 # Install the project into `/app`
 WORKDIR /app
 
-# Enable bytecode compilation
-ENV UV_COMPILE_BYTECODE=1
+# Disable bytecode compilation to avoid EMFILE during build on low nofile limits
```
a7m-1st: It turned out there are at least two kinds of bytecode inhibition:
Local -> container at compile time bytecode transfer, and in container at runtime bytecode generation.

As for the latter one its allowed by default via `ENV PYTHONDONTWRITEBYTECODE=0` (we leave it so if user enabled hot reload for uvicorn)

Image

My only suggestion is to perhaps rename it to differentiate between the two.

```suggestion
# Disable bytecode transfer during compilation to avoid EMFILE during build on low nofile limits
```

a7m-1st: https://stackoverflow.com/questions/59732335/is-there-any-disadvantage-in-using-pythondontwritebytecode-in-docker

server/Dockerfile
a7m-1st, line 47: Btw since we have retry and light installation in line 18, can we combine them both? 
At least we can add the retries here to, wdyt @fengju0213 ?

a7m-1st:
LGTM thanks @Wendong-Fan

server/Dockerfile
fengju0213: good idea! thanks @a7m-1st

fengju0213:
thanks @a7m-1st  and @Wendong-Fan ！

PR #954, opened 2026-01-19
fix after pr-873:update model platform

LuoPengcheng12138:
### Description
Follow-up to #873
Based on the current research, only the `openai_model` and `azure_openai_model` support the `parallel_tool_calls` parameter. For `openai_compatible_model`,`openrouter`,`litellm`, although it may has a consistent interface, whether `parallel_tool_calls` takes effect depends solely on the specific model. Therefore, I will add a judgment for `openai_compatible_model` to prevent calls to the official OpenAI models and official Azure OpenAI models via the compatible interface. As for the conflict issue, it will be resolved by https://github.com/camel-ai/camel/pull/3706.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


LuoPengcheng12138:
hi , @nitpicker55555 Should the OpenRouter and LiteLLM models in this PR (https://github.com/camel-ai/camel/pull/3706) be updated to prevent potential calls to the genuine OpenAI models?

PR #968, opened 2026-01-19
[BUG] Fix List of tools per agent isn't rendering properly (#957)

### LINKED ISSUE #957: [BUG] List of tools per agent isn't rendering properly

0.0.80

### System information

windows 11

### Problem description

The list of tools per agent isn't rendering properly:

Image

### log

Nan

### Additional context

Nan

eureka0928:
## Root Causes Identified

1. **Incomplete CSS class** in `node.tsx:423` - trailing `text-` causing rendering issues
2. **Missing fallback** in `ToolSelect.tsx:633` - badges showed empty when `name` and `mcp_name` were undefined
3. **Inconsistent tool mapping** in `AddWorker/index.tsx` - edit mode used `tool.name` while create mode used `tool.key || tool.mcp_name`

Closes #957 

{C727E786-4242-436B-84D4-8726028E8131}

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


fury0928:
@Wendong-Fan quick contribution for eigent
cc: @a7m-1st

fury0928:
Hi @bytecraftii it seems you assigned to me, not to you, 😉

src/components/AddWorker/index.tsx
```diff
@@ -282,7 +282,7 @@ export function AddWorker({
 						name: workerName,
 						type: workerName as AgentNameType,
 						log: [],
-						tools: [...selectedTools.map((tool) => tool.name)],
+						tools: [...selectedTools.map((tool) => tool.name || tool.key || tool.mcp_name || "")],
```
bytecii: Should we align as above? such as 
```suggestion
						tools: [...selectedTools.map((tool) => tool.name || tool.mcp_name || tool.key || "Unknown Tool")],
```

bytecii: Also will it be problematic if we have multiple "Unknown Tool"?

bytecraftii (commented):
Thanks for the contribution. Left some comments

src/components/AddWorker/index.tsx
fury0928: I think it will be problematic.. can't distinguish which tool is which

In authStore.ts:139, the filter uses string equality:  worker.tools?.filter((t) => t !== tool)                                                                                                     
Removing one "Unknown Tool" would remove all tools named "Unknown Tool". 

What is your opinion? @bytecraftii

Wendong-Fan: would use `tool_${tool.id}` better? also for the current implementation using "Unknown Tool", we should update this to make it unique

fury0928: @Wendong-Fan would you review tihs again? I updated to tool_${tool.id}, actually that is better.
"Unknown Tool" would be buggy

a7m-1st:
Perfect thanks @eureka928, @bytecraftii & @Wendong-Fan !!

PR #975, opened 2026-01-20
fix: terminal toolkit env setting

Wendong-Fan:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


Wendong-Fan:
not perfect yet, would keep enhancing

PR #977, opened 2026-01-20
chore：add kimi provider

LuoPengcheng12138:
### Description
add kimi provider

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/lib/llm.ts
+++ b/src/lib/llm.ts
@@ -77,7 +77,7 @@ export const INIT_PROVODERS: Provider[] = [
     id: 'moonshot',
     name: 'Moonshot',
     apiKey: '',
-    apiHost: 'https://api.moonshot.cn/v1',
+    apiHost: 'https://api.moonshot.ai/v1',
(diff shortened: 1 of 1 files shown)


src/lib/llm.ts
```diff
@@ -73,6 +73,15 @@ export const INIT_PROVODERS: Provider[] = [
     is_valid: false,
     model_type: ""
   },
+  {
+    id: 'moonshot',
+    name: 'Moonshot',
+    apiKey: '',
+    apiHost: 'https://api.moonshot.cn/v1',
```
fengju0213: i think we should use ttps://api.moonshot.ai/v1 
image

fengju0213 (changes requested):

src/lib/llm.ts
LuoPengcheng12138: It‘s OK for now.

PR #767, opened 2025-11-25
feat: Stream mode task spliting

fengju0213:
### What is the purpose of this pull request? 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

Issue #87, opened 2025-08-06
[Feature Request] Task-running front-end supports streaming output


Pakchoioioi:
### Motivation

All information unrelated to structured output or tool invocation should be changed to streaming output.

PR #999, opened 2026-01-21
refactor: extract event_loop_utils module and improve agent response handling

MkDev11:
## Summary

Refactors event loop utilities and agent streaming response handling for improved code organization and thread-safety.

### New Module: `event_loop_utils.py`
- Extracted `set_main_event_loop()` and `_schedule_async_task()` into a dedicated utility module
- Provides thread-safe async task scheduling from worker threads using contextvars + global fallback

### Refactored [listen_chat_agent.py](cci:7://file:///root/74/silver/eigent/backend/app/agent/listen_chat_agent.py:0:0-0:0)
- Added reusable helper methods:
  - [_send_agent_deactivate()](cci:1://file:///root/74/silver/eigent/backend/app/agent/listen_chat_agent.py:113:4-133:9) - Sends agent deactivation events
  - [_extract_tokens()](cci:1://file:///root/74/silver/eigent/backend/app/agent/listen_chat_agent.py:135:4-148:48) - Extracts token count from responses
  - [_stream_chunks()](cci:1://file:///root/74/silver/eigent/backend/app/agent/listen_chat_agent.py:150:4-173:74) / [_astream_chunks()](cci:1://file:///root/74/silver/eigent/backend/app/agent/listen_chat_agent.py:175:4-196:74) - Wraps streaming responses for consistent handling
- Fixed thread-safety: Uses `_schedule_async_task` instead of `asyncio.create_task` for toolkit activation/deactivation calls
- Fixed `AsyncStreamingChatAgentResponse` handling to properly iterate through chunks

### Improved `single_agent_worker.py`
- Better streaming response handling with `last_chunk` tracking
- Properly captures usage info from streaming responses after iteration

### Cleanup
- Removed duplicate code

### DIFF
--- a/backend/app/agent/listen_chat_agent.py
+++ b/backend/app/agent/listen_chat_agent.py
@@ -215,7 +215,7 @@ class ListenChatAgent(ChatAgent):
         response_format: type[BaseModel] | None = None,
     ) -> ChatAgentResponse | StreamingChatAgentResponse:
         task_lock = get_task_lock(self.api_task_id)
-        asyncio.create_task(
+        _schedule_async_task(
             task_lock.put_queue(
                 ActionActivateAgentData(
                     data={
@@ -250,7 +250,7 @@ class ListenChatAgent(ChatAgent):
             if "Budget has been exceeded" in str(e):
                 message = "Budget has been exceeded"
                 logger.warning(f"Agent {self.agent_name} budget exceeded")
-                asyncio.create_task(
+                _schedule_async_task(
                     task_lock.put_queue(ActionBudgetNotEnough())
                 )
             else:
@@ -288,7 +288,7 @@ class ListenChatAgent(ChatAgent):
 
         assert message is not None
 
-        asyncio.create_task(
+        _schedule_async_task(
             task_lock.put_queue(
                 ActionDeactivateAgentData(
                     data={


mkdev5:
@a7m-1st @fengju0213 could you please review the changes and let me know your feedback? thanks!

a7m-1st:
Thanks for the PR @MkDev11 , I don't think I can manage to test this by today (UTC +3).
But have you taken a look at #767 ? kind of similar.

But UI wise can you attach a screenshot of which part you are exactly streaming? Just to get a faster picture !

mkdev5:
@a7m-1st  I added some changes and updated the description. This PR adds some real-time feedback while tasks are running.

Before, you'd just see the task start and then wait until it finished. Now you'll see messages like `[Tool] Shell Exec...` pop up as the agent works, so you know what's actually happening behind the scenes.

I originally wanted to stream the model's actual thinking text (like `ChatGPT` does), but hit a wall - the `camel` library we use crashes when you try to stream with tools enabled. Since pretty much every agent uses tools, I pivoted to showing tool activity instead. Not quite what issue #87 asked for, but it's still a nice improvement.

mkdev5:
streaming messages

a7m-1st:
I see awesome then Thanks @MkDev11, I need time so far as I am little occupied with other tasks. I will catch up with this PR once I got the chance.

mkdev5:
Cool!

a7m-1st:
Hi there @Wendong-Fan , I think this would be a cool feature maybe Douglas can confirm if the UI meets expectations?

fengju0213:
@MkDev11  thanks for your contribution，i noticed that you mentioned that the camel library throws AsyncChatCompletionStreamManager errors when streaming with tools，however, Camel normally supports tool calls in streaming mode. Could I please tell you under what circumstances you encounter the error?

mkdev5:
> @MkDev11 thanks for your contribution，i noticed that you mentioned that the camel library throws AsyncChatCompletionStreamManager errors when streaming with tools，however, Camel normally supports tool calls in streaming mode. Could I please tell you under what circumstances you encounter the error?

Hey @fengju0213, good catch - my comment in the code was a bit misleading. It's not streaming + tools that's the issue, it's streaming + `response_format` (structured output).

Here's what happens:

```python
model = ModelFactory.create(
    model_platform=ModelPlatformType.OPENAI,
    model_type=ModelType.GPT_4O_MINI,
    model_config_dict={'stream': True}
)
agent = ChatAgent(system_message='...', model=model)

# this breaks
response = await agent.astep('query', response_format=SomePydanticModel)
async for chunk in response:
    print(chunk)
```

You get:
```
AttributeError: 'AsyncChatCompletionStreamManager' object has no attribute 'choices'
```

The problem is in `_handle_batch_response` (chat_agent.py:3743) - it tries to do `for choice in response.choices` but OpenAI returns an `AsyncChatCompletionStreamManager` when you combine streaming with structured output, and that object doesn't have `.choices`.

So I just disable streaming when we need structured output:
```python
worker_agent.model_backend.model_config_dict["stream"] = False
response = await worker_agent.astep(prompt, response_format=TaskResult)
```

Let me know what you think

src/store/chatStore.ts
```diff
@@ -840,6 +847,24 @@ const chatStore = (initial?: Partial<ChatStore>) => createStore<ChatStore>()(
 						return;
 					}
 
+					// Handle streaming agent output during task execution
+					if (agentMessages.step === "streaming_agent_output") {
+						const data = agentMessages.data as { process_task_id?: string; content?: string; is_final?: boolean };
+						const { process_task_id, content, is_final } = data;
```
a7m-1st: Lets simplify this to the second line only. Need to add the { process_task_id?: string; content?: string; is_final?: boolean }; to the AgentMessage type itself

src/store/chatStore.ts
```diff
@@ -840,6 +847,24 @@ const chatStore = (initial?: Partial<ChatStore>) => createStore<ChatStore>()(
...
+						
+						if (!process_task_id) return;
+						
+						if (is_final) {
+							// Clear streaming output when final marker received
+							clearStreamingAgentOutput(currentId, process_task_id);
+						} else if (content) {
+							// Append streaming content
+							updateStreamingAgentOutput(currentId, process_task_id, content);
+						}
```
a7m-1st: Can use currentTaskId from the sse loop
```suggestion
						
						if (!process_task_id) return;
						
						if (is_final) {
							// Clear streaming output when final marker received
							clearStreamingAgentOutput(currentTaskId, process_task_id);
						} else if (content) {
							// Append streaming content
							updateStreamingAgentOutput(currentTaskId, process_task_id, content);
						}
```

backend/app/utils/single_agent_worker.py
```diff
@@ -124,20 +131,32 @@ async def _process_task(self, task: Task, dependencies: list[Task]) -> TaskState
                 )
             else:
                 # Use native structured output if supported
-                response = await worker_agent.astep(prompt, response_format=TaskResult)
+                # NOTE: Temporarily disable streaming for structured output because
+                # the camel library doesn't properly handle AsyncChatCompletionStreamManager
+                # from OpenAI's structured output streaming API
+                original_stream = worker_agent.model_backend.model_config_dict.get("stream", False)
+                worker_agent.model_backend.model_config_dict["stream"] = False
```
a7m-1st: Hopfully we an resolve this in https://github.com/camel-ai/camel/pull/3743

backend/app/utils/agent.py
```diff
@@ -705,6 +834,10 @@ def clone(self, with_memory: bool = False) -> ChatAgent:
 
         new_agent.process_task_id = self.process_task_id
 
+        # Preserve model_config_dict (including stream setting) from original agent
+        if hasattr(self, 'model_backend') and hasattr(self.model_backend, 'model_config_dict'):
+            new_agent.model_backend.model_config_dict.update(self.model_backend.model_config_dict)
+
         # Copy memory if requested
```
a7m-1st: We need to deep copy `model_config_dict` if its nested to ensure isolation

backend/app/utils/agent.py
```diff
@@ -269,7 +270,22 @@ def _stream_with_deactivate():
                             last_response = chunk
                             # Accumulate content from each chunk (delta mode)
                             if chunk.msg and chunk.msg.content:
-                                accumulated_content += chunk.msg.content
+                                delta_content = chunk.msg.content
+                                accumulated_content += delta_content
+                                # Stream output chunk to frontend (non-blocking)
+                                asyncio.create_task(
+                                    task_lock.put_queue(
+                                        ActionStreamingAgentOutputData(
```
a7m-1st: Better to use _schedule_async_task to avoid async race conditions in generative functions

a7m-1st:
But feature wise I think it is giving unnecessary overhead to the stream process. To begin with, it is also streaming the status before the agents appear in UI. And also taking a closer look; the user knows the exact status of the toolkit via notice and tools use:

```json
{"step": "notice", "data": {"notice": "I am creating the 'index.html' file with the necessary structure and styling for the FPS game.", "process_task_id": "1769207359549-5780.5"}}	


{"step": "activate_toolkit", "data": {"agent_name": "developer_agent", "process_task_id": "1769207359549-5780.5", "toolkit_name": "Terminal Toolkit", "method_name": "shell exec", "message": "timeout=60, command='npx surge ./fps_game fps-game-jimi.surge.sh', block=False, id='surge_deploy'"}}	
```

I think the activate_toolkit event already gives enough information to be honest compared to:
```
{"step": "streaming_agent_output", "data": {"agent_name": "developer_agent", "process_task_id": "1769207359549-5780.5", "agent_id": "", "content": "[Tool] Shell Exec...\n", "is_final": false}}	
```
Although its a cool feature, I think we can find a better use for it. What do you think @MkDev11 ?

Screenshot 2026-01-24 013348

a7m-1st:
By before starting I mean those:
image

mkdev5:
@a7m-1st please review the changes

fengju0213:
> > @MkDev11 thanks for your contribution，i noticed that you mentioned that the camel library throws AsyncChatCompletionStreamManager errors when streaming with tools，however, Camel normally supports tool calls in streaming mode. Could I please tell you under what circumstances you encounter the error?
> 
> Hey @fengju0213, good catch - my comment in the code was a bit misleading. It's not streaming + tools that's the issue, it's streaming + `response_format` (structured output).
> 
> Here's what happens:
> 
> ```python
> model = ModelFactory.create(
>     model_platform=ModelPlatformType.OPENAI,
>     model_type=ModelType.GPT_4O_MINI,
>     model_config_dict={'stream': True}
> )
> agent = ChatAgent(system_message='...', model=model)
> 
> # this breaks
> response = await agent.astep('query', response_format=SomePydanticModel)
> async for chunk in response:
>     print(chunk)
> ```
> 
> You get:
> 
> ```
> AttributeError: 'AsyncChatCompletionStreamManager' object has no attribute 'choices'
> ```
> 
> The problem is in `_handle_batch_response` (chat_agent.py:3743) - it tries to do `for choice in response.choices` but OpenAI returns an `AsyncChatCompletionStreamManager` when you combine streaming with structured output, and that object doesn't have `.choices`.
> 
> So I just disable streaming when we need structured output:
> 
> ```python
> worker_agent.model_backend.model_config_dict["stream"] = False
> response = await worker_agent.astep(prompt, response_format=TaskResult)
> ```
> 
> Let me know what you think

thank you very much！i will investigate this issue

src/components/WorkFlow/node.tsx
Douglasymlai, line 1: Hi @MkDev11, thanks for contributing this feature! I’d suggest changing leading-tight to leading-normal, or removing the explicit leading setting altogether, to improve readability for users.

mkdev5: Done! please review it again @Douglasymlai

a7m-1st:
> Hopfully we an resolve this in [camel-ai/camel#3743](https://github.com/camel-ai/camel/pull/3743)

I believe we can simplify the configs now bcz it has been fixed @MkDev11 or are they unrelated? i.e. the manual setting of steam mode and on clone too

mkdev5:
> > Hopfully we an resolve this in [camel-ai/camel#3743](https://github.com/camel-ai/camel/pull/3743)
> 
> I believe we can simplify the configs now bcz it has been fixed @MkDev11 or are they unrelated? i.e. the manual setting of steam mode and on clone too

yep, I've simplified it! Removed the workarounds from both `agent.py` and `single_agent_worker.py`, and updated camel-ai to 0.2.85 which includes the fix. The stream mode settings and clone preservation are intentional config, not workarounds, so those stay.

mkdev5:
@a7m-1st please review the changes again

a7m-1st:
All right sure, let me run it through the debugger to reconfirm the flow.
Aside than this, if @Wendong-Fan and @Douglasymlai find it a cool feature, I think it should be good 👍

MkDev11:
@Wendong-Fan @a7m-1st can you let me know your feedback?

MkDev11:
hello @Wendong-Fan can you review the changes again and let me know your feedback? I am still waiting for you

mkdev5:
hello @Wendong-Fan I am really sorry for tagging you, could you please give me any update for me?

a7m-1st:
Hi there @MkDev11 , I have just pushed my enhancements, apologies for the huge delays.

- simplified the set_main_event_loop & _schedule_async_task to not be redundant
- isolate functions to _send_streaming_chunk, _send_agent_deactivate, _extract_tokens & _stream_chunks

Wendong-Fan:
added commit https://github.com/eigent-ai/eigent/pull/999/commits/f7ab0f83c88a30349760fd530e3bb615ad4a5fae  based on review

a7m-1st:
Thanks @Wendong-Fan for your review, and thanks @MkDev11 for the PR

PR #1023, opened 2026-01-22
Use GitHub runner

4pmtong:
### Description

use github runner instead of self host runner

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


.github/workflows/build-view.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: [self-hosted, macOS, ARM64]
+          - os: macos-latest
             arch: arm64
```
a7m-1st, line 18: why was it self-hosted previously anyways?

Wendong-Fan: Hey @a7m-1st, the GitHub runner only has 14 GB of disk space, which wasn’t enough to package our current version. We’ve now reduced the package size so it fits within the limit

Wendong-Fan:
close this pr as was updated in another pr

PR #1056, opened 2026-01-23
:bug: Prebuilt terminal venv found but Python executable missing

4pmtong:
### Description

This PR fixes an issue where the prebuilt terminal venv is found but the Python executable is missing, causing the application to fall back to creating a new terminal venv from scratch. This fallback process is extremely slow as it requires re-downloading Python and all dependency packages.

When this happens, users experience significant delays during startup as Python and all dependencies need to be downloaded and installed again.

Solution:
Replace the missing Python executable with a prebuilt Python binary to avoid the lengthy re-download and installation process.

```
[info] [MacOS] Applied native visual effects
[info] Window content loaded, starting dependency check immediately...
[info] Window shown after content loaded
[info] Window is ready, processing queued protocol URLs...
[info] Using prebuilt binary: /Applications/Eigent.app/Contents/Resources/prebuilt/bin/uv
[info] Using prebuilt binary: /Applications/Eigent.app/Contents/Resources/prebuilt/bin/bun
[info] [DEPS INSTALL] Prebuilt dependencies found, skipping installation
[info] [DEPS INSTALL] Using prebuilt dependencies, creating version file
[info] [DEPS INSTALL] version file updated { currentVersion: '0.0.80' }
[warn] Prebuilt terminal venv found but Python executable missing at: /Applications/Eigent.app/Contents/Resources/prebuilt/terminal_venv/bin/python. Falling back to user terminal venv.
[info] [DEPS INSTALL] Creating terminal base venv (not prebuilt)...
[info] Using prebuilt binary: /Applications/Eigent.app/Contents/Resources/prebuilt/bin/uv
[warn] Prebuilt terminal venv found but Python executable missing at: /Applications/Eigent.app/Contents/Resources/prebuilt/terminal_venv/bin/python. Falling back to user terminal venv.
[info] [DEPS INSTALL] Creating terminal base venv...
```

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


electron/main/utils/process.ts
```diff
@@ -191,10 +241,42 @@ export function getPrebuiltTerminalVenvPath(): string | null {
         log.info(`Using prebuilt terminal venv: ${prebuiltTerminalVenvPath}`);
         return prebuiltTerminalVenvPath;
       } else {
+        // Try to fix the missing Python executable by creating a symlink to prebuilt Python
         log.warn(
           `Prebuilt terminal venv found but Python executable missing at: ${pythonExePath}. ` +
-            `Falling back to user terminal venv.`
+            `Attempting to fix...`
```
fengju0213, line 247: trying to create/repair terminal_venv/bin/python under process.resourcesPath. Could this be caused by missing write permissions?

electron/main/utils/process.ts
```diff
@@ -191,10 +241,42 @@ export function getPrebuiltTerminalVenvPath(): string | null {
...
+              fs.mkdirSync(binDir, { recursive: true });
+            }
+
+            // Create symlink to prebuilt Python
+            if (fs.existsSync(pythonExePath)) {
+              // Remove existing broken symlink or file
+              fs.unlinkSync(pythonExePath);
+            }
+
+            // Calculate relative path for symlink
```
fengju0213, line 268: Would this cause permission issues on Windows?

electron/main/utils/process.ts
```diff
@@ -191,10 +241,42 @@ export function getPrebuiltTerminalVenvPath(): string | null {
...
+            const binDir = isWindows
+              ? path.join(prebuiltTerminalVenvPath, 'Scripts')
+              : path.join(prebuiltTerminalVenvPath, 'bin');
+
+            // Ensure bin directory exists
+            if (!fs.existsSync(binDir)) {
+              fs.mkdirSync(binDir, { recursive: true });
+            }
+
+            // Create symlink to prebuilt Python
```
fengju0213, line 262: If pythonExePath is a dangling symlink, fs.existsSync returns false so we never unlink it, and fs.symlinkSync can throw EEXIST, leaving the repair broken.

electron/main/utils/process.ts
4pmtong: The app bundle includes a prebuilt Python, so when the prebuilt terminal venv's Python executable is missing, we can use the prebuilt Python instead of downloading a new one. This should be handled during the build process to ensure the terminal venv correctly references the bundled Python.

electron/main/utils/process.ts
4pmtong: Good catch. We now handle this during the build process in `before-sign.cjs`, which cleans invalid symlinks before packaging.

electron/main/utils/process.ts
4pmtong: There may be a risk of permission issues on Windows, I'm working on it. we'll enhance Windows permission handling in the next release.

PR #1064, opened 2026-01-23
fix: batch decompose_text events in cloud sync to reduce API spam

### LINKED ISSUE #807: [BUG] error message when running backend
### Motivation

would got error below, but task could executed normally

[error] BACKEND: 2025-12-29 09:14:06,079 - sync_step - ERROR - Failed to sync step to https://dev.eigent.ai/api/chat/steps: ConnectTimeout:

[error] BACKEND: /Users/enrei/.eigent/venvs/backend-0.0.73/lib/python3.10/site-packages/camel/utils/commons.py:1016: RuntimeWarning: coroutine 'TaskLock.put_queue' was never awaited
  return func(*args, **kwargs)

[error] BACKEND: /Users/enrei/.local/share/uv/python/cpython-3.10.16-macos-aarch64-none/lib/python3.10/concurrent/futures/thread.py:85: RuntimeWarning: coroutine 'TaskLock.put_queue' was never awaited
  del work_item

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #1078: [Feature Request] Stream printing for task splitting is too slow for task replay
### Motivation

after we finished the execution of one task, we can replay the task, but it's quiet slow for stream printing part of task splitting as shown in video

https://github.com/user-attachments/assets/fa65ea3a-2899-469f-a154-6c5d85db8aaf

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

bittoby:
Closes #807
Closes #1078

### Description

Optimizes cloud sync by batching high-frequency `decompose_text` events instead of sending them individually, reducing API calls while preserving replay functionality.

### Problem
- `decompose_text` events fire 50-200+ times per task (word-by-word streaming)
- Caused database bloat, API overload, and slow replay (#1078)

### Solution
- **Batch `decompose_text` events** — accumulates text and syncs every ~5 words instead of every token
- **Auto-flush on other events** — buffer flushes before `decompose_progress`, `end`, etc. to preserve event order and prevent data loss
- **~80% fewer API calls** — reduces database bloat and protects rate limits
- **Faster replay** — fewer events to process

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/utils/server/sync_step.py
+++ b/backend/app/utils/server/sync_step.py
@@ -1,3 +1,17 @@
+# ========= Copyright 2025-2026 @ Eigent.ai All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2025-2026 @ Eigent.ai All Rights Reserved. =========
+
 """
 Cloud sync step decorator.
 
@@ -91,7 +105,7 @@ def _buffer_text(task_id: str, content: str):
     """Accumulate decompose_text content in buffer."""
     if task_id not in _text_buffers:
         _text_buffers[task_id] = ""
-    _text_buffers[task_id] = content
+    _text_buffers[task_id] += content
 
 
 def _should_flush(task_id: str) -> bool:
@@ -134,13 +148,19 @@ def _parse_value(value):
 def _get_task_id(args):
     if not args or not hasattr(args[0], "task_id"):
         return None
-    
+
     chat = args[0]
     task_lock = get_task_lock_if_exists(chat.project_id)
-    
+
     if task_lock and getattr(task_lock, "current_task_id", None):
         return task_lock.current_task_id
-    
+
+    if not task_lock:
+        logger.warning(
+            f"Task lock not found for project_id {chat.project_id}, "
+            f"using chat.task_id"
+        )
+
     return chat.task_id
 
 
@@ -148,5 +168,5 @@ async def _send(url, data):
     try:
         async with httpx.AsyncClient(timeout=5.0) as client:
             await client.post(url, json=data)
-    except Exception:
-        pass
+    except Exception as e:
+        logger.error(f"Failed to sync step to {url}: {type(e).__name__}: {e}")


Crystora:
@Wendong-Fan Could you please review this PR?

Crystora:
@4pmtong Could you please review this PR? Thanks

a7m-1st:
HI there @bittoby thanks for the PR, I will review by tmrw max, but one thing I am kind of confused, why are we disabling the stream? I don't think its optional as the whole history mechanism depends on it.

If disabled, the chat history will be lost

a7m-1st:
> No authentication token was included

The /chat/steps endpoint doesn't require authentication actually. 

```
async def create_chat_step(step: ChatStepIn, session: Session = Depends(session)):
```

I suggest you analyze why it is failing instead, is it due to rate limits? or api overload due to the decompose_text event?

backend/app/utils/server/sync_step.py
```diff
@@ -1,96 +1,112 @@
...
-                    logger.warning(f"Task lock not found for project_id {chat.project_id}, using chat.task_id")
-                    task_id = chat.task_id
-
-            if task_id:
-                # TODO: Filter out unnecessary events to avoid database bloat
-                # - Skip "decompose_text" streaming events (sent 50-200+ times per task)
-                # - Only sync structural events: decompose_progress, task_state, create_agent, etc.
-                # - Consider batching or deduplication for high-frequency events
-                # - Extract and add task dependencies for analytics
-
```
a7m-1st, line 73: We should attend to this flow instead @bittoby

Crystora:
@a7m-1st 
Thanks for your feedback! Sorry for the confusion.
I thought cloud sync and SSE streaming were the same thing. When I saw the ConnectTimeout error, I thought disabling cloud sync would fix it without side effects. I was wrong.

I've updated the code so cloud sync is enabled by default again and added filtering for decompose_text events to reduce API spam.  I also removed the auth token since the endpoint doesn't require it.
I really appreciate you can review again.

Crystora:
For batching/deduplication and task dependencies, I'm skipping these for now because:
Filtering decompose_text already removes the bulk of API spam. The remaining events are low-frequency, so batching would add complexity without much benefit.
Let me know if you'd like me to implement these, or if the current fix is enough for now. Thanks!

a7m-1st:
HI there @bittoby , thanks for the updates and following my suggestions. It is true that the token and batching would not fix the issue, but I think skipping the event would simply not show the decomposed subtasks on replay of chat, or am I mistaken?

Honestly what I suggest is kind of batch the 'decompose_text' event by sending X words (e.g. 5 words) at the same time rather than word by word or char by char. This will also fix this issue #1078 where replay is too slow. 
I think this way we can keep the PR small, protect rate limits and replay faser.  What do you think? @bittoby 

cc: @LuoPengcheng12138 @Wendong-Fan

Crystora:
Thanks for the feedback! @a7m-1st 
You’re right - if we skip `decompose_text` completely, replay will break because the subtask text won’t get saved to the cloud DB.
I’ll update the PR so `decompose_text` gets sent in batches (like collecting ~5 words before syncing). The buffer will also auto-flush when other events come in (like `decompose_progress` or end), so we won’t lose anything.
This should cut down the API calls a lot and still keep replay working, just faster.

I’ll push the updated changes soon.

a7m-1st:
All right perfect! Looking forwards to it

Crystora:
@a7m-1st I updated all. Please review again.

backend/app/utils/server/sync_step.py
```diff
@@ -1,96 +1,152 @@
...
-                    )
-                )
-            yield value
 
-    return wrapper
+def _buffer_text(task_id: str, content: str):
+    """Accumulate decompose_text content in buffer."""
+    if task_id not in _text_buffers:
+        _text_buffers[task_id] = ""
+    _text_buffers[task_id] = content
```
Wendong-Fan: i think here should be `_text_buffers[task_id] += content`?

backend/app/utils/server/sync_step.py
```diff
@@ -1,96 +1,152 @@
...
-            res = await client.post(url, json=data)
-            # logger.info(res)
-        except Exception as e:
-            logger.error(f"Failed to sync step to {url}: {type(e).__name__}: {e}")
+async def _send(url, data):
+    try:
+        async with httpx.AsyncClient(timeout=5.0) as client:
+            await client.post(url, json=data)
+    except Exception:
+        pass
```
Wendong-Fan: better add log instead of slientily ignore

Crystora:
Thank you for your feedback @Wendong-Fan 
If you have any headache issues, feel free to reach me out anytime. I’m always on this project :)

PR #1076, opened 2026-01-24
enhance: always render html in iframe

### LINKED ISSUE #1012: [Feature Request] HTML Rendering Still Not Working as Expected on macOS (Apple Silicon)
### Motivation

We are still seeing unexpected behavior with HTML rendering.

**Environment**
OS: macOS M
App version: 0.0.80

**Issue**
HTML content is not rendered as expected in this version.
The problem persists even after testing with the latest packaged build.

Image

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #1051: [Feature Request] Enhance current agent folder html display
### Motivation

the generated html file shown in the agent folder is not in good format compared with opened locally, we need to further enhance the display

sceenshot from agent folder:
Image

same file opened in browser:

Image

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

a7m-1st:
### Description
- fixes #1051 
- fixes #1012 

- removed the optional rendering
- removed translation
- isolated html sanitization incase of revert

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/lib/htmlSanitization.ts
+++ b/src/lib/htmlSanitization.ts
@@ -19,16 +19,16 @@ import DOMPurify, { Config } from 'dompurify';
  * These should be blocked even in sandboxed iframes as a defense-in-depth measure.
  */
 export const DANGEROUS_PATTERNS = [
-  /ipcRenderer/gi,
-  /window\s*\[\s*['"`]ipcRenderer['"`]\s*\]/gi,
-  /parent\s*\.\s*ipcRenderer/gi,
-  /top\s*\.\s*ipcRenderer/gi,
-  /frames\s*\[\s*\d+\s*\]\s*\.\s*ipcRenderer/gi,
-  /require\s*\(\s*['"`]electron['"`]\s*\)/gi,
-  /process\s*\.\s*versions\s*\.\s*electron/gi,
-  /nodeIntegration/gi,
-  /webSecurity/gi,
-  /contextIsolation/gi,
+  /ipcRenderer/i,
+  /window\s*\[\s*['"`]ipcRenderer['"`]\s*\]/i,
+  /parent\s*\.\s*ipcRenderer/i,
+  /top\s*\.\s*ipcRenderer/i,
+  /frames\s*\[\s*\d+\s*\]\s*\.\s*ipcRenderer/i,
+  /require\s*\(\s*['"`]electron['"`]\s*\)/i,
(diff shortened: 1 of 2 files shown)


src/lib/htmlSanitization.ts
```diff
@@ -0,0 +1,157 @@
...
+  /nodeIntegration/gi,
+  /webSecurity/gi,
+  /contextIsolation/gi,
+];
+
+/**
+ * Check if HTML content contains dangerous patterns that could attempt
+ * to access Electron/Node.js APIs.
+ */
+export function containsDangerousContent(html: string): boolean {
```
Wendong-Fan, line 38: this was not used in HtmlRenderer.

src/lib/htmlSanitization.ts
```diff
@@ -0,0 +1,157 @@
...
+  /ipcRenderer/gi,
+  /window\s*\[\s*['"`]ipcRenderer['"`]\s*\]/gi,
+  /parent\s*\.\s*ipcRenderer/gi,
+  /top\s*\.\s*ipcRenderer/gi,
+  /frames\s*\[\s*\d+\s*\]\s*\.\s*ipcRenderer/gi,
+  /require\s*\(\s*['"`]electron['"`]\s*\)/gi,
+  /process\s*\.\s*versions\s*\.\s*electron/gi,
+  /nodeIntegration/gi,
+  /webSecurity/gi,
+  /contextIsolation/gi,
```
Wendong-Fan: The g flag causes .test() to maintain lastIndex state, which leads to alternating true/false results as below
```
  const re = /ipcRenderer/gi;
  re.test("ipcRenderer"); // true
  re.test("ipcRenderer"); // false (!)
```
  This means every second malicious file would bypass the security check, keep only i for case-insensitive matching would be better?

src/lib/htmlSanitization.ts
a7m-1st: oh yes, I isolated it just in case we need to revert it

src/lib/htmlSanitization.ts
a7m-1st: all right, didn't focus into it honestly, just refactored what was present 🆙 
Thanks for the enhancement !

PR #1088, opened 2026-01-26
Enhance schedule trigger

Pakchoioioi:
I search on some competitor schedule trigger, and did some enhancement related to schedule feature.  
- Set triggers ad Once, daily, weekly and monthly Schedule
- add expire date (optional)
- delete schedule summary, it doesn't make any sense

Screenshot 2026-01-26 at 20 09 59
Screenshot 2026-01-26 at 20 09 38
Screenshot 2026-01-26 at 20 09 24
Screenshot 2026-01-26 at 20 09 16

To @Douglasymlai : Please optimize the visual components and handle language adaptation/localization.
To @a7m-1st : I need your help debugging the backend. (Please note: I have not made any changes to the backend code myself).

When debugging, please pay special attention to the following requirements:

Time Validation: The selected time must be later than the current server time; otherwise, the action cannot be saved.

Weekly "Select Weekdays": 
1. Must support toggle selection (click to select/deselect). 
2. Must support multi-selection (e.g., selecting both Tuesday and Wednesday). 
3. At least one day must be selected.

Monthly Special Logic:
Needs to handle short months and leap years. If a user selects the 31st, but the scheduled month only has 30 days (or 28/29 days), the default behavior is not to trigger.

Trigger Expiration Mechanism:
When the Current Date > Expiration Date, the trigger must automatically become invalid or stop executing.

### DIFF
--- a/src/components/Trigger/SchedulePicker.tsx
+++ b/src/components/Trigger/SchedulePicker.tsx
@@ -12,9 +12,14 @@ import { useTranslation } from "react-i18next";
 
 type FrequencyType = "once" | "daily" | "weekly" | "monthly";
 
+export type ScheduleData = {
+    date?: string; // ISO date string for one-time execution (YYYY-MM-DD) - needed because cron has no year
+    expirationDate?: string; // ISO date string for when recurring schedule expires
+};
+
 type SchedulePickerProps = {
-    value: string; // cron expression or schedule data
-    onChange: (cronExpression: string) => void;
+    value: string; // JSON stringified ScheduleData
+    onChange: (scheduleData: string) => void;
     onValidationChange?: (isValid: boolean) => void;
 };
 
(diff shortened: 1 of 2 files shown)


a7m-1st:
> To @Douglasymlai : Please optimize the visual components and handle language adaptation/localization. To @a7m-1st : I need your help debugging the backend. (Please note: I have not made any changes to the backend code myself).
> 
> When debugging, please pay special attention to the following requirements:
> 
> Time Validation: The selected time must be later than the current server time; otherwise, the action cannot be saved.
> 
> Weekly "Select Weekdays":
> 
> 1. Must support toggle selection (click to select/deselect).
> 2. Must support multi-selection (e.g., selecting both Tuesday and Wednesday).
> 3. At least one day must be selected.
> 
> Monthly Special Logic: Needs to handle short months and leap years. If a user selects the 31st, but the scheduled month only has 30 days (or 28/29 days), the default behavior is not to trigger.
> 
> Trigger Expiration Mechanism: When the Current Date > Expiration Date, the trigger must automatically become invalid or stop executing.

Thanks @Pakchoioioi for the PR, yes I believe we have addressed all those suggestions UI wise thanks to @Douglasymlai in #1060 , and as for the logic:
a. Regarding monthly special logic and time validation: if occuring at intervals, then we don't need to validate if first excecution will be missed, I think its upto the user's responsibility if too late. As for "once" to be run in a minute, we can add frontend logic only for this case.

b. Trigger Expiration: in the latest server PR expiration logic + max fails configs are added.

tldr:
I think we can close this PR for now since its implemented in a different PR, thanks for your suggestions and enhancements @Pakchoioioi 🙌

PR #1091, opened 2026-01-27
refactor: fix note taking toolkit minor bug and refactor toolkit listen

bytecii:
### Description

- Remove the default `.md` working directory as it's a (potential) bug and for now we all specify working directory from agents
- Refactor `backend/app/utils/listen/toolkit_listen.py`
- Add unit tests for them

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/backend/app/utils/listen/toolkit_listen.py
+++ b/backend/app/utils/listen/toolkit_listen.py
@@ -67,21 +67,17 @@ def _format_result(
 ) -> str:
     """Format the result or error as a string."""
     if error is not None:
-        return str(error)
+        return _truncate(str(error))
 
     if return_msg_formatter is not None:
-        return return_msg_formatter(res)
+        return _truncate(return_msg_formatter(res))
 
     if isinstance(res, str):
-        return res
+        return _truncate(res)
 
     try:
-        # Notice:
-        # JSON-serializable data is kept intact for DB storage and replay
-        return json.dumps(res, ensure_ascii=False)
+        return _truncate(json.dumps(res, ensure_ascii=False))
     except TypeError:
-        # Notice:
-        # Non-JSON data (random objects) is truncated as it's less useful now
         return _truncate(str(res))
 
 
--- a/backend/tests/app/utils/listen/test_toolkit_listen.py
+++ b/backend/tests/app/utils/listen/test_toolkit_listen.py
@@ -142,6 +142,40 @@ def test_format_result_with_non_serializable():
     assert "object" in result
 
 
+@pytest.mark.unit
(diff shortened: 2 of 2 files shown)


backend/app/utils/toolkit/note_taking_toolkit.py
```diff
@@ -34,9 +29,16 @@ def __init__(
         working_directory: str | None = None,
```
Zephyroam, line 29: It's better to remove the default value `None` if `working_directory` is required.

backend/app/utils/listen/toolkit_listen.py
```diff
@@ -13,31 +13,157 @@
...
+    if return_msg_formatter is not None:
+        return return_msg_formatter(res)
+
+    if isinstance(res, str):
+        return res
+
+    try:
+        return json.dumps(res, ensure_ascii=False)
+    except TypeError:
+        return _truncate(str(res))
```
Zephyroam, line 81: Why does only this line call `_truncate`?

backend/app/utils/listen/toolkit_listen.py
```diff
@@ -13,31 +13,157 @@
...
+from app.service.task import (ActionActivateToolkitData,
+                              ActionDeactivateToolkitData, get_task_lock,
+                              process_task)
 from app.utils.toolkit.abstract_toolkit import AbstractToolkit
-from app.service.task import process_task
-import logging
 
 logger = logging.getLogger("toolkit_listen")
 
+MAX_LENGTH = 500
```
Zephyroam, line 32: consider making it configurable?

bytecii: I think for now we can keep the original one as it's not customer facing?

backend/app/utils/toolkit/note_taking_toolkit.py
bytecii: I think the main reason is that the parent camel `NoteTakingToolkit` supports the `None`
Now keeping `None` mainly because of matching the parent signature. Let me add a todo to revisit this later

backend/app/utils/listen/toolkit_listen.py
bytecii: I think the main reason is that JSON-serializable data will be used in the postgres stored as jsonb for future query and replay but the non json data is usually some tools output and it's less useful for now (so that we can truncate)
Added notice comments to them

Wendong-Fan (commented):
thanks @bytecii and @Zephyroam !

PR #1094, opened 2026-01-27
enhance: enhance cdp

fengju0213:
- Problem: CDP port occupancy is tracked in a global pool and workforce shutdown
    clears the entire pool, causing parallel tasks to reuse the same port; clone
    temporarily mutates the parent toolkit’s CDP config without guaranteed restoration
    and leaks pre-acquired ports on failure; cdp_browsers uses a mutable default and
    can leak state across requests.
- Fix: Track session/task ownership in the CDP pool and release by task instead of
  clearing globally; wrap clone mutation in try/finally with cleanup on failure;
  switch cdp_browsers to Field(default_factory=list).

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/agent/factory/browser.py
+++ b/backend/app/agent/factory/browser.py
@@ -42,16 +42,22 @@ class CdpBrowserPoolManager:
 
     def __init__(self):
         self._occupied_ports: dict[int, str] = {}
+        self._session_to_port: dict[str, int] = {}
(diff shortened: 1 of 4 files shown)


nitpicker55555:
Thanks @fengju0213 !

PR #1098, opened 2026-01-27
fix(cookie_manager): robustness — None-safe value, search_cookies domain, temp DB cleanup

sharpenteeth:
## Summary
- get_cookies_for_domain: handle None row value to avoid TypeError when Chrome has NULL cookie values
- search_cookies: guard domain when None to avoid AttributeError on .lower()
- _get_cookies_connection: clean up temp DB file on connect failure
- Any type hint for cookie domain dicts


sharpenteeth:
Hi, @Wendong-Fan.
Could you review my PR,please?
Thanks

PR #1101, opened 2026-01-28
chore: fix mcp is called synchronously.

fengju0213:

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/store/chatStore.ts
+++ b/src/store/chatStore.ts
@@ -693,11 +693,6 @@ const chatStore = (initial?: Partial<ChatStore>) =>
         lockedTaskId = newTaskId;
       };
 
-      let mcpLocal = {};
-      if (window.ipcRenderer) {
-        mcpLocal = await window.ipcRenderer.invoke('mcp-list');
-      }
-
       fetchEventSource(api, {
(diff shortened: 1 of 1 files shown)


src/store/chatStore.ts
```diff
@@ -717,7 +722,10 @@ const chatStore = (initial?: Partial<ChatStore>) =>
               api_key: apiModel.api_key,
               api_url: apiModel.api_url,
               extra_params: apiModel.extra_params,
-              installed_mcp: { mcpServers: {} },
+              installed_mcp:
+                mcpLocal && Object.keys(mcpLocal).length > 0
+                  ? mcpLocal
+                  : { mcpServers: {} },
```
a7m-1st: Thanks @fengju0213, just an unrelated issue where here we are passing the .eigent/mcp.json but the mcp page isn't reflecting this file.

a7m-1st: Image
Just reverted it for now as its better to focus on server synched config.

fengju0213:
@a7m-1st thanks！

PR #1102, opened 2026-01-28
chore: fix replay when self host

fengju0213:
## 🐞 Root Cause

The local Electron app always **overwrote `SERVER_URL`** when spawning the backend,
forcing it to: https://dev.eigent.ai/api

As a result:

- Step events were **synced to the cloud** instead of the local server
- The local server had **no step data**
- Local replay **got stuck with no steps to replay**

---

## ✅ Fix

Updated the Electron backend launcher to:

- Read `SERVER_URL` from `~/.eigent/.env` **or existing environment variables**
- **Only fall back to the cloud URL** if `SERVER_URL` is not provided

After restarting the app and running a new task:

- Step events are **stored locally**
- Local replay **works as expected**

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/init.ts
+++ b/electron/main/init.ts
@@ -36,38 +36,50 @@ import {
 
 const execAsync = promisify(exec);
 
-const DEFAULT_SERVER_URL = "https://dev.eigent.ai/api";
+const DEFAULT_SERVER_URL = 'https://dev.eigent.ai/api';
 
 function readEnvValue(filePath: string, key: string): string | undefined {
-    try {
-        if (!fs.existsSync(filePath)) return undefined;
-        const content = fs.readFileSync(filePath, 'utf-8');
-        const lines = content.split(/\r?\n/);
-        const line = lines.find((l) => {
-            const trimmed = l.trim();
-            return trimmed && !trimmed.startsWith('#') && trimmed.startsWith(`${key}=`);
-        });
-        if (!line) return undefined;
-        let value = line.trim().slice(key.length + 1).trim();
-        // Strip surrounding quotes (single or double)
-        if ((value.startsWith('"') && value.endsWith('"')) ||
-            (value.startsWith("'") && value.endsWith("'"))) {
-            value = value.slice(1, -1);
-        }
-        return value;
-    } catch (error) {
-        log.warn(`Failed to read ${key} from ${filePath}:`, error);
-        return undefined;
+  try {
+    if (!fs.existsSync(filePath)) return undefined;
+    const content = fs.readFileSync(filePath, 'utf-8');
+    const lines = content.split(/\r?\n/);
+    const line = lines.find((l) => {
+      let trimmed = l.trim();
+      if (!trimmed || trimmed.startsWith('#')) return false;
+      // Handle lines with 'export' prefix (e.g. export SERVER_URL=value)
+      if (trimmed.startsWith('export ')) {
+        trimmed = trimmed.slice(7).trim();
+      }
+      return trimmed.startsWith(`${key}=`);
+    });
+    if (!line) return undefined;
+    let raw = line.trim();
+    // Strip 'export ' prefix before extracting value
+    if (raw.startsWith('export ')) {
+      raw = raw.slice(7).trim();
     }
+    let value = raw.slice(key.length + 1).trim();
+    // Strip surrounding quotes (single or double)
+    if (
+      (value.startsWith('"') && value.endsWith('"')) ||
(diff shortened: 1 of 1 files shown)


electron/main/init.ts
```diff
@@ -25,6 +26,29 @@ import { PromiseReturnType } from "./install-deps";
...
 const execAsync = promisify(exec);
 
+const DEFAULT_SERVER_URL = "https://dev.eigent.ai/api";
+
+function readEnvValue(filePath: string, key: string): string | undefined {
+    try {
+        if (!fs.existsSync(filePath)) return undefined;
+        const content = fs.readFileSync(filePath, 'utf-8');
+        const lines = content.split(/\r?\n/);
+        const line = lines.find((l) => l.trim() && !l.trim().startsWith('#') && l.startsWith(`${key}=`));
```
Wendong-Fan: Small inconsistency here: `l.trim()` is used for the empty/comment checks, but `l.startsWith(...)` runs on the raw (untrimmed) line. So if someone has a leading space like `  SERVER_URL=xxx` in their .env, the trim checks pass but the key match fails silently.

Suggestion: store `l.trim()` in a variable and use it for all checks:
```ts
const line = lines.find((l) => {
    const trimmed = l.trim();
    return trimmed && !trimmed.startsWith('#') && trimmed.startsWith(`${key}=`);
});
```

electron/main/init.ts
```diff
@@ -25,6 +26,29 @@ import { PromiseReturnType } from "./install-deps";
...
+        if (!line) return undefined;
+        return line.slice(key.length + 1).trim();
```
Wendong-Fan: This will break if the .env value is quoted, e.g. `SERVER_URL="https://my-server.com/api"`. The returned value would include the surrounding quotes, making the URL invalid.

Worth adding a strip for single/double quotes after the trim:
```ts
let value = line.trim().slice(key.length + 1).trim();
if ((value.startsWith('"') && value.endsWith('"')) ||
    (value.startsWith("'") && value.endsWith("'"))) {
    value = value.slice(1, -1);
}
return value;
```

(Also note: since we changed to matching on `trimmed`, need `line.trim().slice(...)` here too.)

electron/main/init.ts
```diff
@@ -25,6 +26,29 @@ import { PromiseReturnType } from "./install-deps";
...
+        log.warn(`Failed to read ${key} from ${filePath}:`, error);
+        return undefined;
+    }
+}
+
+function buildLocalServerUrl(proxyUrl: string | undefined): string | undefined {
+    if (!proxyUrl) return undefined;
+    const trimmed = proxyUrl.trim().replace(/\/+$/, '');
+    if (!trimmed) return undefined;
+    return `${trimmed}/api`;
```
Wendong-Fan: If someone already has `/api` in their proxy URL (e.g. `VITE_PROXY_URL=http://localhost:8000/api`), this would produce `http://localhost:8000/api/api`.

Quick guard before appending:
```ts
if (trimmed.endsWith('/api')) return trimmed;
return `${trimmed}/api`;
```

electron/main/init.ts
```diff
@@ -174,10 +198,49 @@ export async function startBackend(setPort?: (port: number) => void): Promise<an
...
+    let resolvedSource = 'default';
+
+    if (envProxyEnabled) {
+        resolvedServerUrl = buildLocalServerUrl(envProxyUrl);
+        resolvedSource = 'process.env VITE_*';
+    }
+
+    const devServerUrl = process.env.VITE_DEV_SERVER_URL;
+    if (!resolvedServerUrl && devServerUrl) {
+        const devEnvPath = path.join(process.cwd(), '.env.development');
```
Wendong-Fan: `process.cwd()` is unreliable in packaged Electron apps. It varies depending on how the user launched the app (Finder, terminal, Spotlight, etc.). Since this block only runs when `VITE_DEV_SERVER_URL` is set (i.e. during dev), `app.getAppPath()` would be a safer choice here and still point to the project root.

Wendong-Fan (commented):
Nice fix! The root cause analysis is solid and the multi-source resolution approach makes sense. Left a few inline comments on edge cases that could bite us in production.

Wendong-Fan (changes requested):
added commit https://github.com/eigent-ai/eigent/pull/1102/changes/67e42f5b0261ea3fa1114b5738874fedf86cf7fa

fengju0213:
thanks! @Wendong-Fan

PR #1103, opened 2026-01-28
 Fix symlink traversal in task uploads

LuoPengcheng12138:
### Description
  This PR closes a symlink-based escape in task file collection and upload:

  - Task file listing now skips symlinks and validates real paths to ensure they stay inside the working directory.
  - Electron’s file enumeration uses lstat + realpath to prevent recursive traversal through external links.
  - sanitize_command() to validate ln -s will implement in Camel repo

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


electron/main/fileReader.ts
```diff
@@ -548,7 +549,13 @@ export class FileReader {
 				if (this.hiddenFolders.includes(file)) continue;
 
 				const filePath = path.join(dirPath, file);
-				const stats = fs.statSync(filePath);
```
4pmtong: How about add a try-catch here to handle potential errors(e.g  file doesn't exist, permission denied)

LuoPengcheng12138: Okay, but there's already a try-catch block wrapping the entire function, maybe adding another one here seems a bit redundant?

a7m-1st (commented):
Thanks @LuoPengcheng12138 , can you add test cases for this? I have already added a comprehensive test suite for electron file mngmnt. LGTM though 👍

electron/main/fileReader.ts
4pmtong: This really depends on the intended behavior:
Current approach (fail fast): Any file error stops the entire traversal
With inner try-catch (best effort): Skip problematic files and continue
What's your preference here? Should one bad file fail the entire upload, or should we skip it and continue with the rest?

LuoPengcheng12138: You are right~ we should skip it and continue with the rest.

Issue #1082, opened 2026-01-25
[BUG] Maybe remove note taking toolkit


lightaime:
### What version of eigent are you using?

0.0.82

### System information

ProductName:		macOS
ProductVersion:		15.3.1

### Problem description

Some notes created by the `TerminalToolkit` is not able to read by the `NoteTakingToolkit` which create lots of problem. `TerminalToolkit` can already be used for creating notes and read notes if we add some prompts. So `NoteTakingToolkit` is not needed.

In this example, document agent created a `.md` file with terminal toolkit in task 1:
<img width="699" height="702" alt="Image" src="https://github.com/user-attachments/assets/0501a3e5-a869-40a1-be17-4707720d83e6" />

The document agent try to read the notes with note taking toolkit. However this is no note since the note is created using terminal toolkit which is not being registered by the note taking toolkit.
<img width="699" height="702" alt="Image" src="https://github.com/user-attachments/assets/e7fa9207-ce8f-4e2d-a8a5-16db97e8e993" />

### log

The agent also hanged in task No.3 for unknown reason

[eigent-0.0.82-darwin-arm64-1769325865504.log](https://github.com/user-attachments/files/24842798/eigent-0.0.82-darwin-arm64-1769325865504.log)

Wendong-Fan:
as discussed with Guohao, we need to add some prompt instructions. For example, before a worker starts a task, they can use `ls` to check whether there is any important content in the working directory, and they can also read the relevant files; well-designed sys prompt could help model has better performance on using terminal to handle note

bytecii:
Taking look.

PR #1106, opened 2026-01-28
refactor: refactor agent and modify sys prompt for terminal ls

bytecii:
### Description

1. Refactor agents to `app/agent` instead of keeping everything in the `app/utils` More details please refer to https://github.com/eigent-ai/eigent/blob/refactor/refactor_agent/backend/app/agent/README.md
2. Modify the agent system prompt equipped with the terminal toolkit to fix https://github.com/eigent-ai/eigent/issues/1082 by adding
```
- **IMPORTANT:** Before the task gets started, you can use `shell_exec` to
    run `ls {working_directory}` to check for important files in the working
    directory, and then use terminal commands like `cat`, `grep`, or `head`
    to read and examine these files.
```
4. Make the test agent unit tests work and refactor them

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/backend/app/agent/factory/browser.py
+++ b/backend/app/agent/factory/browser.py
@@ -25,6 +25,7 @@ from app.service.task import Agents
 from app.utils.file_utils import get_working_directory
 from app.utils.toolkit.human_toolkit import HumanToolkit
 from app.utils.toolkit.hybrid_browser_toolkit import HybridBrowserToolkit
+# TODO: Remove NoteTakingToolkit and use TerminalToolkit instead
 from app.utils.toolkit.note_taking_toolkit import NoteTakingToolkit
 from app.utils.toolkit.search_toolkit import SearchToolkit
 from app.utils.toolkit.terminal_toolkit import TerminalToolkit
@@ -34,9 +35,8 @@ from camel.toolkits import ToolkitMessageIntegration
 
 def browser_agent(options: Chat):
     working_directory = get_working_directory(options)
-    logger.info(
-        f"Creating browser agent for project: {options.project_id} in directory: {working_directory}"
-    )
+    logger.info(f"Creating browser agent for project: {options.project_id} "
+                f"in directory: {working_directory}")
     message_integration = ToolkitMessageIntegration(
         message_handler=HumanToolkit(
             options.project_id, Agents.browser_agent).send_message_to_user)
--- a/backend/app/agent/factory/social_medium.py
+++ b/backend/app/agent/factory/social_medium.py
@@ -11,8 +11,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ========= Copyright 2025-2026 @ Eigent.ai All Rights Reserved. =========
-from camel.messages import BaseMessage
-
 from app.agent.agent_model import agent_model
 from app.agent.listen_chat_agent import logger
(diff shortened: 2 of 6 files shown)


backend/app/agent/factory/social_medium.py
bytecii, line 1: This name need and will be changed later...

backend/app/agent/factory/browser.py
```diff
@@ -0,0 +1,130 @@
...
+from app.agent.listen_chat_agent import logger
+from app.agent.prompt import BROWSER_SYS_PROMPT
+from app.agent.utils import NOW_STR
+from app.component.environment import env
+from app.model.chat import Chat
+from app.service.task import Agents
+from app.utils.file_utils import get_working_directory
+from app.utils.toolkit.human_toolkit import HumanToolkit
+from app.utils.toolkit.hybrid_browser_toolkit import HybridBrowserToolkit
+from app.utils.toolkit.note_taking_toolkit import NoteTakingToolkit
```
fengju0213, line 29: if we already modified the sys prompt for terminal ，does this mean we no longer need the note toolkit?

backend/app/agent/prompt.py
```diff
@@ -0,0 +1,597 @@
...
+    - Perform comprehensive spreadsheet operations including:
+        * Sheet creation, deletion, and data clearing
+        * Cell-level operations (read, write, find specific values)
+        * Row and column manipulation (add, update, delete)
+        * Range operations for bulk data processing
+        * Data export to CSV format for compatibility
+    - Handle complex data structures with proper formatting and validation
+    - Support for both programmatic data entry and manual cell updates
+
+- Terminal and File System:
```
fengju0213, line 302: i think we can modify the prompt and remove note toolkit？

fengju0213 (commented):
thanks！ @bytecraftii  left some comments below

backend/app/agent/factory/browser.py
bytecii: I think we should still keep the note taking toolkit for now. As now note taking still serve as an explicit and core tool to let agents exchange their works / summaries via file.
Agree that we should gradually remove the notetaking toolkit but we should maybe further benchmark (modifying the system prompt to enable file writing and appending) before change. let me add a todo for now

fengju0213 (approved):
I have no further questions. The existing changes are all good and won't disrupt the original logic.

Wendong-Fan:
hey @bytecii @fengju0213 , do we have follow up issue to remove note taking toolkit? please link with this pr or previous issue https://github.com/eigent-ai/eigent/issues/1082 if we have 

The current fix adds `ls + cat/grep` guidance in system prompts to help agents discover files created via terminal.
  However, this introduces redundancy:

  1. Tool redundancy: NoteTakingToolkit is still registered, giving agents both read_note and terminal commands for reading files
  2. Prompt conflict: System prompts instruct agents to use both read_note and terminal commands, potentially causing
  duplicate operations

bytecii:
@Wendong-Fan there is another PR that may help to mitigate the file issue https://github.com/eigent-ai/eigent/pull/1145
Let's create some example use cases and test it in the benchmark before changing too much https://github.com/eigent-ai/eigent/pull/1138?

PR #1111, opened 2026-01-30
enhance upload to s3

4pmtong:
### Description

After Build and Build-View finish, installers are uploaded to S3 so users can download from S3.
Build (tag release): uploads to releases/v{version}/ and releases/latest/
Build-View (manual test build): uploads to test-builds/{test-id}/
Requires 4 repo secrets: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_S3_BUCKET, AWS_REGION

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


.github/workflows/build.yml
```diff
@@ -260,3 +260,241 @@ jobs:
...
+      - name: Upload to S3 (versioned)
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          VERSION="${{ steps.version.outputs.version }}"
+          BUCKET="${{ secrets.AWS_S3_BUCKET }}"
+
+          echo "Uploading version $VERSION to S3 bucket: $BUCKET"
+
+          # Upload macOS ARM64 files
+          if [ -d "release/mac-arm64" ] && [ "$(ls -A release/mac-arm64)" ]; then
```
bytecii: nit:
```bash
declare -a targets=(
  "release/mac-arm64 mac-arm64"
  "release/mac-intel mac-intel"
  "release/win-x64 win-x64"
  "release/linux-x64 linux-x64"
)
for t in "${targets[@]}"; do
...
```

.github/workflows/build.yml
```diff
@@ -260,3 +260,241 @@ jobs:
...
+              --metadata-directive REPLACE \
+              --cache-control "public, max-age=31536000" \
+              --exclude "*.yml" \
+              --exclude "*.blockmap"
+
+            aws s3 sync release/mac-arm64/ "s3://$BUCKET/releases/v$VERSION/mac-arm64/" \
+              --exclude "*" \
+              --include "*.yml" \
+              --include "*.blockmap" \
+              --cache-control "public, max-age=300"
```
bytecii: Curious why we have a different max age here?

.github/workflows/build.yml
```diff
@@ -260,3 +260,241 @@ jobs:
...
+
+          echo "Uploading version $VERSION to S3 bucket: $BUCKET"
+
+          # Upload macOS ARM64 files
+          if [ -d "release/mac-arm64" ] && [ "$(ls -A release/mac-arm64)" ]; then
+            echo "Uploading macOS ARM64 files..."
+            aws s3 sync release/mac-arm64/ "s3://$BUCKET/releases/v$VERSION/mac-arm64/" \
+              --content-type "binary/octet-stream" \
+              --metadata-directive REPLACE \
+              --cache-control "public, max-age=31536000" \
```
bytecii: Will
```bash
--cache-control "public, max-age=31536000, immutable"
```
better?

.github/workflows/build-view.yml
```diff
@@ -230,3 +230,111 @@ jobs:
...
+      - name: Extract version
+        id: version
+        run: |
+          # Create a version using timestamp for test builds
+          VERSION="test-$(date +%Y%m%d-%H%M%S)"
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Extracted version for test build: $VERSION"
+
+      # Configure AWS credentials
+      - name: Configure AWS credentials
```
Wendong-Fan, line 244: These AWS steps (credentials, upload, URL generation) don't have any conditional guard. If the repo doesn't have AWS secrets configured, the whole `merge-release` job will fail. Consider adding `if: ${{ secrets.AWS_ACCESS_KEY_ID != '' }}` like how build.yml guards its S3 steps.

.github/workflows/build.yml
```diff
@@ -260,3 +260,241 @@ jobs:
...
+          fi
+
+          # Upload Linux files
+          if [ -d "release/linux-x64" ] && [ "$(ls -A release/linux-x64)" ]; then
+            echo "Uploading Linux files..."
+            aws s3 sync release/linux-x64/ "s3://$BUCKET/releases/v$VERSION/linux-x64/" \
+              --content-type "binary/octet-stream" \
+              --metadata-directive REPLACE \
+              --cache-control "public, max-age=31536000" \
+              --exclude "*.yml"
```
Wendong-Fan: The macOS and Windows sections both exclude `*.blockmap` from the long cache sync and include it in the short cache sync, but Linux only handles `*.yml`. Worth adding `--exclude "*.blockmap"` here and `--include "*.blockmap"` in the second sync for consistency. Same for the latest sync section below.

.github/workflows/build.yml
4pmtong: @bytecii long cache for installers, short cache for update metadata.

.github/workflows/build.yml
4pmtong: @Wendong-Fan At line 246 we only move AppImage and yml for Linux — the Linux (AppImage) build doesn’t produce separate `.blockmap` files (blockmap is embedded in the AppImage).

.github/workflows/build.yml
4pmtong: Good idea. Versioned URLs never change. Added to all four versioned sync steps.

.github/workflows/build.yml
4pmtong: Thanks @bytecii, Adopted the loop for the versioned upload. For versioned paths (releases/v$VERSION/...) everything is immutable, so we now use a single sync per platform with max-age=31536000, immutable for all files (including yml/blockmap); no separate short cache for metadata there. Latest upload stays per-platform with two syncs and short cache for yml/blockmap so clients get fresh update info quickly.

.github/workflows/build-view.yml
4pmtong: @Wendong-Fan We already have the AWS secrets configured in this repo. I’ve also added the conditional guard here (aws-check + env-based if and steps.aws-check.outcome) so forks or runs without AWS config won’t fail on these steps.

Issue #839, opened 2026-01-13
[Feature Request] Add windows native appearance


Douglasymlai:
### Motivation

The Windows app’s appearance needed improvement to incorporate native Windows settings and better align with expected visual and user experience standards.

glowsenior:
can I pick up? @Douglasymlai

Douglasymlai:
@glowsenior Yeah, that’d be amazing — thank you. I don’t have a Windows device on hand, so testing on my side is always a bit tricky. I’ll assign the issue to you first and add more context/notes in the description.

Also, if you’re familiar with tweaking Electron apps to feel more native on Windows, feel free to adjust that too. Once you open a PR, I’m happy to review and help test/verify.

glowsenior:
thx,

glowsenior:
Hello @Douglasymlai , I made the PR. Please check that. https://github.com/eigent-ai/eigent/pull/1114

PR #1114, opened 2026-01-31
feat(windows): improve native Windows appearance and theme support

glowsenior:
### Description

This PR improves the Windows app's native appearance and theme support to better align with Windows UX standards. The app now uses native Windows frame controls, respects system dark/light mode, and removes macOS-specific visual effects that don't work well on Windows.

**Key Changes:**
- ✅ Respect Windows system theme (dark/light mode) instead of forcing light theme
- ✅ Use native Windows frame with native window controls
- ✅ Remove macOS-only visual effects on Windows (vibrancy, roundedCorners, transparency)
- ✅ Use solid background on Windows for better native appearance
- ✅ Hide custom window controls on Windows (native frame provides controls)
- ✅ Add Windows system theme change listener for dynamic theme updates

**Files Changed:**
- `electron/main/index.ts` - Platform-specific window configuration and theme handling
- `src/components/WindowControls/index.tsx` - Hide controls on Windows
- `src/components/TopBar/index.tsx` - Hide custom controls on Windows

**Testing:**
- ✅ Tested on Windows with native frame
- ✅ Verified system theme detection (dark/light mode)
- ✅ Confirmed native window controls work properly
- ✅ macOS behavior unchanged (backward compatible)

- [ ] Bug fix
- [X] New Feature (#839 )
- [ ] Documentation update
- [ ] Other

### Additional Context

This addresses the enhancement request to improve Windows app appearance and native integration. The changes ensure Windows users get a more familiar, native experience while maintaining the existing macOS design.

**Before:** Frameless window with custom controls, forced light theme, macOS visual effects
**After:** Native Windows frame, system theme support, platform-appropriate styling

### DIFF
--- a/.github/workflows/build-view.yml
+++ b/.github/workflows/build-view.yml
@@ -230,3 +230,99 @@ jobs:
           else
             find temp-linux-x64 -name "*.AppImage" -exec mv {} release/linux-x64/ \; || true
           fi
+
+      # Extract version for test builds
+      - name: Extract version
+        id: version
+        run: |
+          # Create a version using timestamp for test builds
+          VERSION="test-$(date +%Y%m%d-%H%M%S)"
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Extracted version for test build: $VERSION"
+
+      # Configure AWS credentials (skipped when AWS secrets are not configured)
+      - name: Configure AWS credentials
+        id: aws-check
+        if: env.AWS_ACCESS_KEY_ID != ''
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+
+      # Upload to S3 - test directory
+      - name: Upload to S3 (test build)
+        if: steps.aws-check.outcome == 'success'
+        run: |
+          VERSION="${{ steps.version.outputs.version }}"
+          BUCKET="${{ secrets.AWS_S3_BUCKET }}"
+
+          echo "Uploading test build $VERSION to S3 bucket: $BUCKET"
+
+          declare -a targets=(
+            "release/mac-arm64 mac-arm64"
+            "release/mac-intel mac-intel"
+            "release/win-x64 win-x64"
+            "release/linux-x64 linux-x64"
+          )
+          for t in "${targets[@]}"; do
+            src_dir="${t%% *}"
+            path_suffix="${t##* }"
+            if [ -d "$src_dir" ] && [ "$(ls -A "$src_dir")" ]; then
+              echo "Uploading $path_suffix files..."
+              aws s3 sync "$src_dir/" "s3://$BUCKET/test-builds/$VERSION/$path_suffix/" \
+                --content-type "binary/octet-stream" \
+                --metadata-directive REPLACE \
+                --cache-control "public, max-age=300"
+            fi
+          done
+
+          echo "Test build $VERSION uploaded successfully"
+
+      # Generate download URLs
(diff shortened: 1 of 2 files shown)


electron/main/index.ts
```diff
@@ -1263,6 +1268,7 @@ let installationLock: Promise<PromiseReturnType> = Promise.resolve({
 // ==================== window create ====================
 async function createWindow() {
   const isMac = process.platform === 'darwin';
+  const isWindows = process.platform === 'win32';
```
Wendong-Fan: This redeclares `isWindows` which already exists at module scope (line 171). The module-level `const` is accessible inside this function via closure, so this shadow is unnecessary and a bit confusing. Can just remove this line.

electron/main/index.ts
```diff
@@ -1281,22 +1287,35 @@ async function createWindow() {
...
-    frame: false,
+    // Use native frame on Windows for better native integration
+    frame: isWindows ? true : false,
     show: false, // Don't show until content is ready to avoid white screen
-    transparent: true,
-    vibrancy: 'sidebar',
-    visualEffectState: 'active',
-    backgroundColor: '#f5f5f580',
+    // Only use transparency on macOS (works better there)
+    transparent: isMac ? true : false,
```
Wendong-Fan: Heads up: the original code had `transparent: true` for **all** platforms. By using `isMac ? true : false`, this also turns off transparency on Linux (where `isMac` is also false). That's a regression for Linux users.

Since only Windows needs transparency disabled, this should be:
```js
transparent: !isWindows,
```

electron/main/index.ts
```diff
@@ -1281,22 +1287,35 @@ async function createWindow() {
...
+    // macOS-only visual effects
+    vibrancy: isMac ? 'sidebar' : undefined,
+    visualEffectState: isMac ? 'active' : undefined,
+    // Solid background on Windows, semi-transparent on macOS
+    backgroundColor: isWindows ? '#ffffff' : '#f5f5f580',
```
Wendong-Fan: Since `nativeTheme.themeSource` is set to `'system'` for Windows (line 173), the user could already be in dark mode when the window is created. Hardcoding `#ffffff` here will cause a white flash before the renderer paints its dark content.

Would suggest picking the color dynamically:
```js
backgroundColor: isWindows
  ? (nativeTheme.shouldUseDarkColors ? '#1e1e1e' : '#ffffff')
  : '#f5f5f580',
```

electron/main/index.ts
```diff
@@ -1281,22 +1287,35 @@ async function createWindow() {
...
+    // macOS-specific title bar styling
     titleBarStyle: isMac ? 'hidden' : undefined,
     trafficLightPosition: isMac ? { x: 10, y: 10 } : undefined,
     icon: path.join(VITE_PUBLIC, 'favicon.ico'),
-    roundedCorners: true,
+    // Rounded corners only on macOS
+    roundedCorners: isMac ? true : false,
```
Wendong-Fan: Same issue as `transparent` above. The original had `roundedCorners: true` for all platforms, so using `isMac` here also breaks Linux. Should be:
```js
roundedCorners: !isWindows,
```

electron/main/index.ts
```diff
@@ -1920,6 +1939,20 @@ app.whenReady().then(async () => {
...
 
+  // ==================== Windows theme change listener ====================
+  // Listen for system theme changes on Windows and notify renderer
+  if (isWindows) {
+    nativeTheme.on('updated', () => {
+      const theme = nativeTheme.shouldUseDarkColors ? 'dark' : 'light';
+      log.info(`[THEME] System theme changed to: ${theme}`);
+      // Notify renderer process about theme change
+      if (win && !win.isDestroyed()) {
+        win.webContents.send('system-theme-changed', { theme });
```
Wendong-Fan: This sends `system-theme-changed` to the renderer, but there's no corresponding listener in the preload script or the renderer code anywhere in this PR. The message just goes into the void.

I'd remove this entire theme listener block for now and add it back together with the renderer-side handler when that's ready. Dead code that looks functional tends to confuse people down the road.

Wendong-Fan:
added commit https://github.com/eigent-ai/eigent/pull/1114/commits/bc75e62e193443727ec7579edcc89d6abd605830 based on review

PR #1159, opened 2026-02-05
feat: add open agent folder in IDE from Folder view

### LINKED ISSUE #996: [Feature Request] Support opening the agent folder with IDE like VS code or Cursor
### Motivation

Support opening the agent folder with IDEs like vs code or cursor for editing files or running code. It is help for the scenario that we want to use Eigent for deep research and brainstorming and then cursor for furthur development.

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

eureka0928:
### Description

Add the ability to open an agent's project folder directly in an IDE (VS Code or Cursor) from the Folder view panel. This feature improves developer workflow by providing quick access to agent files.

Closes #996 

**Changes:**
- Add "Open in IDE" button in the Folder view header (terminal icon)
- Add IDE preference setting in General settings (VS Code, Cursor, or System File Manager)
- Implement IPC handlers for getting project folder path and opening in IDE
- Fall back to system file manager if selected IDE is not installed
- Add i18n translations for new UI elements

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -927,8 +927,110 @@ function registerIpcHandlers() {
   ipcMain.handle(
     'open-in-ide',
     async (_event, folderPath: string, ide: string) => {
-      const cmd = ide === 'vscode' ? 'code' : ide === 'cursor' ? 'cursor' : '';
+      const getIDECommand = (): string => {
+        const platform = process.platform;
+        const homeDir = homedir();
+
+        if (ide === 'vscode') {
+          if (platform === 'darwin') {
+            // macOS: Check common VS Code CLI paths
+            const vscodePaths = [
+              '/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code',
+              '/usr/local/bin/code',
+            ];
+            for (const p of vscodePaths) {
+              if (existsSync(p)) return p;
+            }
+            log.warn(
+              '[IDE] VS Code not found on macOS, using system file manager'
+            );
+            return '';
+          } else if (platform === 'win32') {
+            // Windows: Check common VS Code paths
+            const vscodePaths = [
+              path.join(
+                homeDir,
+                'AppData',
+                'Local',
+                'Programs',
+                'Microsoft VS Code',
+                'bin',
+                'code.cmd'
+              ),
+              path.join(
+                homeDir,
+                'AppData',
+                'Local',
+                'Programs',
+                'Microsoft VS Code',
+                'Code.exe'
+              ),
+              'C:\\Program Files\\Microsoft VS Code\\bin\\code.cmd',
+              'C:\\Program Files\\Microsoft VS Code\\Code.exe',
+            ];
+            for (const p of vscodePaths) {
+              if (existsSync(p)) return p;
+            }
+            log.warn(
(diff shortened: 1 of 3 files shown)


fury0928:
@Wendong-Fan @4pmtong @bytecii would you review my PR?
Thank you

Wendong-Fan (commented):
Nice feature! The IDE integration looks solid 

Found a couple of i18n issues:

**Missing translations for other locales**

New keys were only added to `en-us`. we need to add translations to: zh-Hans, zh-Hant, ja, ko, ar, de, es, fr, it, ru

**Hardcoded error message**

In src/components/Folder/index.tsx, the error message should use i18n:

```tsx
// Before
toast.error(result.error || 'Failed to open folder');

// After  
toast.error(result.error || t('chat.failed-to-open-folder'));
```

Add `failed-to-open-folder` key to all chat.json files as well.

fury0928:
@Wendong-Fan what is the next step?

src/components/Folder/index.tsx
```diff
@@ -446,21 +448,55 @@ export default function Folder({ data: _data }: { data?: Agent }) {
...
-              title={isCollapsed ? t('chat.open') : t('chat.close')}
-            >
-              <ChevronsLeft
-                className={`h-6 w-6 text-icon-secondary ${
-                  isCollapsed ? 'rotate-180' : ''
-                } transition-transform ease-in-out`}
-              />
-            </Button>
+            <div className="flex items-center">
+              {!isCollapsed && window.electronAPI?.getProjectFolderPath && (
```
Wendong-Fan: Should also check for `openInIDE` here since the onClick handler uses both methods. Otherwise it could throw a runtime error if `openInIDE` is undefined.

```tsx
{!isCollapsed &&
  window.electronAPI?.getProjectFolderPath &&
  window.electronAPI?.openInIDE && (
```

electron/main/index.ts
```diff
@@ -914,6 +914,57 @@ function registerIpcHandlers() {
...
+      const manager = checkManagerInstance(fileReader, 'FileReader');
+      const result = manager.createProjectStructure(email, projectId);
+      return result.path;
+    }
+  );
+
+  ipcMain.handle(
+    'open-in-ide',
+    async (_event, folderPath: string, ide: string) => {
+      const cmd = ide === 'vscode' ? 'code' : ide === 'cursor' ? 'cursor' : '';
```
Wendong-Fan: On macOS, `cursor` and `code` are not in PATH by default. They live inside the app bundle:

• `/Applications/Cursor.app/Contents/Resources/app/bin/cursor`
• `/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code`

Same issue on Windows where IDEs install to user directories.

Also, with `shell: true`, the shell spawns successfully even when the command doesn't exist, so the spawn event fires and returns success without actually opening anything.

Suggestion: check if the IDE binary exists at known paths first. If not found, fall back to system file manager immediately.

Wendong-Fan:
> @Wendong-Fan what is the next step?

thanks @eureka928 ! I added some updates please feel free to check, now LGTM

PR #1164, opened 2026-02-05
fix: persist subtask edits to backend on add/update/delete (#930)

### LINKED ISSUE #930: [Feature Request] Subtask Create / Update / Delete Status Is Not Persisted
### Motivation

When editing subtasks, changes are not being saved correctly. The following issues occur:
**Subtask state is not persisted**
- Deleted subtasks reappear after reassignment
- Newly added subtasks disappear
- Modified subtask status is not saved

**Edit and Save buttons on the right panel are not working**
- Clicking Edit has no effect
- Clicking Save does not persist any changes

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

eureka0928:
## Summary

- Persist subtask edits (update, delete) to the backend via the existing `PUT /task/{project_id}` endpoint, so user changes survive page refresh and task reassignment
- Debounce keystroke-driven updates (500ms) while sending delete actions immediately
- Fix the Save button in `TaskItem` to explicitly call `onUpdate` with current content before exiting edit mode
- Cancel any pending debounced persist when a task starts, preventing stale data from overwriting the confirmed task list

Closes #930 

## Root Cause

`updateTaskInfo()`, `deleteTaskInfo()`, and the Save button in the frontend only updated local Zustand state — no backend call was made. This meant subtask edits were lost on page refresh, task failure, or reassignment.

## Changes

| File | Change |
|------|--------|
| `src/store/chatStore.ts` | Add `persistSubtaskEdits()` helper with debounce support; call it from `updateTaskInfo` (debounced) and `deleteTaskInfo` (immediate); add `cancelPendingSubtaskPersist()` called before task start to avoid stale writes |
| `src/components/ChatBox/TaskBox/TaskItem.tsx` | Fix Save button `onClick` to explicitly call `onUpdate(taskInfo.content)` before exiting edit mode |

## Test plan

- [ ] Edit a subtask's content and click Save — verify the change persists after page refresh
- [ ] Edit a subtask and wait 500ms without saving — verify the debounced PUT fires (check Network tab)
- [ ] Delete a subtask — verify deletion persists after page refresh
- [ ] Add a new subtask, type content — verify it persists after page refresh
- [ ] Start a task immediately after editing subtasks — verify the confirmed list is not overwritten by a stale debounced call
- [ ] Run `npm run type-check` — no new errors introduced

### DIFF
--- a/src/components/ChatBox/TaskBox/TaskItem.tsx
+++ b/src/components/ChatBox/TaskBox/TaskItem.tsx
@@ -91,6 +91,7 @@ export function TaskItem({
             onChange={(e) => onUpdate(e.target.value)}
             onBlur={() => {
               setTimeout(() => {
+                onSave();
                 setIsFocus(false);
               }, 100);
             }}


src/components/ChatBox/TaskBox/TaskItem.tsx
```diff
@@ -114,7 +114,11 @@ export function TaskItem({
             </Button>
           ) : (
             <Button
-              onClick={(e) => handleFocus(e, false)}
+              onClick={(e) => {
+                // Explicitly persist current content on Save click
+                onUpdate(taskInfo.content);
```
Wendong-Fan: This still goes through the 500ms debounce since `updateTaskInfo` always passes `debounce=true` to `persistSubtaskEdits`.

Maybe add an `immediate` param to `updateTaskInfo` so clicking Save actually saves right away?

src/store/chatStore.ts
```diff
@@ -2863,6 +2913,9 @@ const chatStore = (initial?: Partial<ChatStore>) =>
         i === index ? { ...item, content } : item
       );
       setTaskInfo(activeTaskId, targetTaskInfo);
+
+      // Persist to backend with debounce (fires on every keystroke)
+      persistSubtaskEdits(targetTaskInfo, true);
```
Wendong-Fan: nit: small typo in comment, "keystroke" reads a bit technical. Also this always uses debounce, but if we want Save button to be instant, we might need a param here.

fury0928:
Looks good?

fury0928:
HI @4pmtong would you review this as well?
Thank you

Wendong-Fan:
hey @eureka928 , i noticed there's some unexpected behaviour need to be fixed, the added task content would be brought to the modified task, also there's some warning msg as shown in the recorded video below



https://github.com/user-attachments/assets/b818d4f5-e18f-4c24-ac9c-ab5629e1dc00

src/store/chatStore.ts
```diff
@@ -2870,6 +2920,9 @@ const chatStore = (initial?: Partial<ChatStore>) =>
...
```
nitpicker55555: This implementation would cause any key press in editing task to be sent to the backend in real time. In other words, while the user is editing a task card, an API request would be triggered every 500 ms, which makes the “Save changes” button meaningless and results in multiple unnecessary backend communications during every key press.
@Wendong-Fan do we need this effect?

fury0928:
@Wendong-Fan @nitpicker55555 would you review again?
I addressed all issues

Wendong-Fan (approved):
thanks @eureka928 and @nitpicker55555 ! added one minor update https://github.com/eigent-ai/eigent/pull/1164/commits/abf3fe941a015fa2eeaab2e167a809755cc32a55 , now LGTM

PR #1168, opened 2026-02-06
Fix/1158 task lock not found

### LINKED ISSUE #1158: [BUG] Task lock not found

0.0.8.2

### System information

windows10 desktop application

### Problem description

Image
The old task ended,but new task had no reaction.

### log

[error] BACKEND: 2026-02-05 18:17:43,676 - chat_controller - INFO - Chat improvement requested
2026-02-05 18:17:43,676 - task_service - ERROR - Task lock not found
[error] BACKEND: [31mERROR[0m:    Exception in ASGI application

### Additional context

_No response_

nj-tools:
Closes #1158

### Description

Fixes “task lock not found” errors by making lock acquisition/release idempotent and improving retry handling. This prevents failed runs when a lock record is missing or already released.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/controller/chat_controller.py
+++ b/backend/app/controller/chat_controller.py
@@ -21,7 +21,7 @@ from pathlib import Path
 
 from dotenv import load_dotenv
 from fastapi import APIRouter, Request, Response
-from fastapi.responses import StreamingResponse
+from fastapi.responses import JSONResponse, StreamingResponse
 
 from app.component import code
 from app.component.environment import sanitize_env_path, set_user_env_path
@@ -49,6 +49,7 @@ from app.service.task import (
     delete_task_lock,
     get_or_create_task_lock,
     get_task_lock,
+    get_task_lock_if_exists,
     set_current_task_id,
     task_locks,
 )
@@ -256,7 +257,25 @@ def improve(id: str, data: SupplementChat):
         "Chat improvement requested",
         extra={"task_id": id, "question_length": len(data.question)},
     )
-    task_lock = get_or_create_task_lock(id)
+    task_lock = get_task_lock_if_exists(id)
+
+    if task_lock is None:
+        # SSE session no longer exists (disconnected, timed out, or stopped).
+        # Return 410 Gone so the frontend can fall back to POST /chat which
+        # creates both a new task lock AND a new SSE consumer.
+        chat_logger.warning(
+            "Task lock not found for improve request, "
+            "returning 410 so client reconnects via POST /chat",
+            extra={"project_id": id},
+        )
+        return JSONResponse(
+            status_code=410,
+            content={
+                "code": 410,
+                "error": "session_expired",
+                "message": "Session expired. Please reconnect.",
+            },
+        )
 
     # Allow continuing conversation even after task is done
     # This supports multi-turn conversation after complex task completion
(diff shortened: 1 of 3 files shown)


src/components/ChatBox/index.tsx
bytecii, line 595: Should we do something like this?
```
if (res?.code === 410) {
  // Re-establish SSE + task lock via POST /chat
  await chatStore.startTask(nextTaskId, ...);
}
```

bytecii (commented):
Maybe we can also add some unit tests for this. Thanks!

src/components/ChatBox/index.tsx
nj-tools: I think we need that because when the improve endpoint returns 410, we fall back to startTask() which reestablishes both the SSE consumer and the task lock.

nj-tools:
> Maybe we can also add some unit tests for this. Thanks!

Yep, for sure. I added test cases for this.

backend/app/controller/chat_controller.py
nitpicker55555, line 379: Why we still use get_task_lock in supplement, stop, add_task

nj-tools: We keep get_task_lock() on supplement/stop/add_task because they only make sense with an existing active session; creating a new lock there would silently drop the action (no SSE consumer). improve is user-facing, so we return 410 and let the client reconnect via POST /chat to re-create both lock + SSE.

nitpicker55555 (commented):
The new warning log uses `project_id`:
```python
extra={"project_id": id}
```
But the existing info log 3 lines above uses `task_id` for the same `id` parameter:
```python
extra={"task_id": id, "question_length": len(data.question)}
```
Since `id` here is actually a project ID (from the route `/chat/{id}` and frontend's `projectStore.activeProjectId`), recommend unifying to `project_id` in both places.

backend/app/controller/chat_controller.py
```diff
@@ -256,7 +257,25 @@ def improve(id: str, data: SupplementChat):
...
+
+    if task_lock is None:
+        # SSE session no longer exists (disconnected, timed out, or stopped).
+        # Return 410 Gone so the frontend can fall back to POST /chat which
+        # creates both a new task lock AND a new SSE consumer.
+        chat_logger.warning(
+            "Task lock not found for improve request, "
+            "returning 410 so client reconnects via POST /chat",
+            extra={"project_id": id},
+        )
```
nitpicker55555, line 270: 258 line extra={"task_id": id
but 269 line extra={"project_id": id

this id should be task_id right?

src/components/ChatBox/index.tsx
nitpicker55555, line 595: The backend now returns 410 when the task lock is missing, but currently there's no frontend handling for it. The existing fetchPost call in ChatBox/index.tsx is fire-and-forget (no await, no .then()/.catch()), and handleResponse in src/api/http.ts doesn't check HTTP status codes — so the 410 response gets silently discarded. Could we add corresponding frontend handling (e.g. await the response, detect 410, and show a reconnect prompt to the user) to make this fix actually take effect?

bytecii:
@darkmoon-o can you resolve the conflicts and fix according to the comments so that I can re take a look?

nj-tools:
> @darkmoon-o can you resolve the conflicts and fix according to the comments so that I can re take a look?

Okay, I will check.

PR #1173, opened 2026-02-07
Fix EPIPE crash on shutdown and llvmlite LLVM version mismatch

wrsfmss:
## Summary
- Fix uncaught EPIPE errors that crash the app during shutdown when electron-log's console transport writes to a closed stdout pipe
- Fix llvmlite build failure caused by cmake finding LLVM 14 instead of the required LLVM 20, and uv's build isolation stripping the `CMAKE_ARGS` env var

### EPIPE Fix (`electron/main/index.ts`)
- Added stream error handlers on `process.stdout`/`process.stderr` to silently catch broken pipe writes
- Added `uncaughtException` handler that ignores EPIPE errors
- Disabled console log transport in `window-all-closed` and `before-quit` handlers before window destruction

### llvmlite/LLVM Fix (`electron/main/utils/process.ts` + `electron/main/install-deps.ts`)
- `getUvEnv()` discovers LLVM 20 and sets `CMAKE_ARGS` with `-DLLVM_DIR` and `LLVM_CONFIG`
- Enhanced PATH to include LLVM 20 bin directories
- Added `--no-build-isolation-package llvmlite` to `uv sync` args so `CMAKE_ARGS` reaches the build subprocess

## Test plan
- [x] App launches without EPIPE crash
- [x] All 218 backend packages install successfully (including llvmlite)
- [x] Backend starts on port 5001 with all routers loaded
- [x] Health check passes
- [x] Zero EPIPE errors in logs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -25,6 +25,17 @@ import {
   shell,
 } from 'electron';
 import log from 'electron-log';
+
+// ==================== EPIPE error prevention ====================
+// During app shutdown, the stdout/stderr pipes may close before all log
+// writes complete. This causes uncaught EPIPE errors that crash the app.
+// Handle them gracefully by ignoring broken pipe writes.
+for (const stream of [process.stdout, process.stderr]) {
+  stream?.on?.('error', (err: NodeJS.ErrnoException) => {
+    if (err.code === 'EPIPE') return; // Silently ignore broken pipe
+  });
+}
+
 import FormData from 'form-data';
 import fsp from 'fs/promises';
 import mime from 'mime';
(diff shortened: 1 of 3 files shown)


Wendong-Fan (commented):
Hey @wrsfmss, Thanks for investigating this issue!

I found that `unstructured` was actually an unused dependency in our backend (no imports anywhere in the codebase). I've removed it in #1177, which also eliminates the entire `llvmlite`/`numba` dependency chain, so the LLVM build issue is resolved.

the **EPIPE fix** in this PR still looks valuable and is unrelated to the dependency issue. Would you mind updating this PR to keep just the EPIPE crash fix and remove the llvmlite/LLVM changes? That would make it easier to review and merge.

Wendong-Fan:
closing this PR as it has been inactive

Issue #959, opened 2026-01-19
[Feature Request] Verify that the Firecrawl MCP is functioning properly


Pakchoioioi:
### Motivation

Background: Models fail to call firecrawl MCP, and doesn't recognize it, UI displays it is successfully implemented but models doesn't see firecrawl tools.

To do:
We need to check whether the Firecrawl MCP is viable.
Check whether models can recognize custom MCPs.

fengju0213:
@Pakchoioioi  i have already investigated this issue ，firecrawl mcp works well in my side，Is it because the user didn't add MCP to the agent?

fengju0213:
maybe we can let the user see this document.https://docs.eigent.ai/core/workers

Crystora:
@Wendong-Fan @Pakchoioioi I am interested in this issue. Can I pick this up?

Wendong-Fan:
thanks @bittoby ! issue assigned

Wendong-Fan:
hey @bittoby , any updates or anything we could support?

Crystora:
@Wendong-Fan I am doing that now. I will send pr in a few hours

PR #1194, opened 2026-02-09
chore: remove unused mcpLocal code in chatStore

bittoby:
### Description
Fixes models not seeing or using installed MCPs (e.g. Firecrawl). The frontend was loading local MCP config from Electron but always sending an empty `installed_mcp` in the chat request, so the backend never received the real MCP list. The request body now sends the actual local MCP config when starting a chat so the backend can connect to MCP servers and expose their tools to the model.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/pages/History.tsx
+++ b/src/pages/History.tsx
@@ -132,8 +132,9 @@ export default function Home() {
         />
       </div>
       {/* Navbar */}
+      {/* -top-px avoids a visible hairline: at top-0 subpixel rounding can leave a gap; */}
       <div
-        className={`sticky top-0 z-20 flex flex-col items-center justify-between border-x-0 border-t-0 border-solid border-border-disabled bg-bg-page-default px-20 pb-4 pt-10`}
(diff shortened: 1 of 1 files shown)


Crystora:
@Wendong-Fan @Pakchoioioi Please review this PR.

Crystora:
@Wendong-Fan could you merge this pr?

Wendong-Fan:
Hi @bittoby, after investigating, I found that MCP functionality actually works correctly through the AddWorker path when users add an agent via AddWorker and select MCP tools, the `new_agents[].mcp_tools` is properly passed to the backend and processed

  As @fengju0213 mentioned in the https://github.com/eigent-ai/eigent/issues/959#issuecomment-3771959266, the original issue might be a user configuration problem rather than a code bug, a cleaner approach might be to remove the unused mcpLocal code instead since it's loaded but never used, this would improve code quality without changing the intended behavior, i will update directly in your PR, WDYT?

PR #1205, opened 2026-02-10
Fix add llvmlite to mac intel

4pmtong:
### Description
reopen

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/electron/main/utils/process.ts
+++ b/electron/main/utils/process.ts
@@ -772,6 +772,7 @@ export function getPrebuiltTerminalVenvPath(): string | null {
 
   if (needsFix) {
     fixPyvenvCfgPlaceholder(pyvenvCfgPath);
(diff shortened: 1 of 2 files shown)


4pmtong:
Can test installers here https://github.com/eigent-ai/eigent/actions/runs/21827026566

electron/main/utils/process.ts
fengju0213, line 739: I have verified this locally using a packaged installation, and everything starts up normally in my environment.there is still a potential risk: getPrebuiltTerminalVenvPath writes .terminal_venv_fixed under
 process.resourcesPath, which may be read-only in some install/permission setups. These scenarios were not fully covered in my local testing.
Would it be better to write the marker to a user-writable location and wrap marker read/write in try/catch, so that marker write failures only log a warning and don’t interrupt dependency checks or fallback behavior?

PR #1214, opened 2026-02-10
Fix Duplicate task execution after network reconnection

### LINKED ISSUE #1212: [Feature Request] Duplicate task execution after network reconnection and system wake-up
### Motivation

After a task completes and the computer is put into sleep mode (or the network is disconnected), the same task triggers and runs for a second time automatically once the system wakes up and the connection is restored.

https://github.com/user-attachments/assets/76540651-9f66-4699-9fbb-b86c73694350

[eigent-0.0.82-darwin-arm64-1770736813398.log](https://github.com/user-attachments/files/25214163/eigent-0.0.82-darwin-arm64-1770736813398.log)

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

spider-yamet:
### Description

Fixes duplicate task execution when the SSE connection is lost and then restored (e.g. network reconnect, Wi‑Fi disconnect/reconnect, or system sleep/wake). Previously, `fetch-event-source` would retry on connection errors even when the task had already finished, which could re-run the same task.

**What changed**
- In `chatStore.ts`, the SSE `onerror` handler now checks if the task status is `FINISHED` before allowing a retry. If the task is already finished, it logs that retry is being stopped to avoid duplicate execution, cleans up the connection, and throws so `fetch-event-source` does not retry.
- Added a unit test in `chatStore.test.ts`: "SSE onerror - no retry when task already finished (issue #1212)" to ensure `fetchEventSource` is only called once and the "already finished, stopping retry" path is exercised when the task is `FINISHED` and an error occurs.

**Verification**
- Run: `npm run test -- --run test/unit/store/chatStore.test.ts -t "issue #1212"`

Closes #1212

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/app/agent/toolkit/hybrid_browser_toolkit.py
+++ b/backend/app/agent/toolkit/hybrid_browser_toolkit.py
@@ -599,6 +599,17 @@ class HybridBrowserToolkit(BaseHybridBrowserToolkit, AbstractToolkit):
         # Use typing_extensions.TypedDict for Pydantic <3.12 compatibility.
         return await super().browser_sheet_input(cells=cells)
 
+    def get_tools(self):
+        tools = super().get_tools()
+        for tool in tools:
+            if not getattr(tool.func, "__listen_toolkit__", False):
+                cls_method = getattr(type(self), tool.func.__name__, None)
+                if cls_method and getattr(
+                    cls_method, "__listen_toolkit__", False
+                ):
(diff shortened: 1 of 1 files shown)


spider-yamet:
@Wendong-Fan @4pmtong Could you please review my PR?

PR #1221, opened 2026-02-11
Feat skills

4pmtong:
#### Summary

Introduces a **Skills System** and adds an **Agents** tab to the main navigation.

#### Skills System

- **Backend**: SkillToolkit with user config and agent access control; integrated into all agents (Developer, Browser, Document, Multi Modal, Social Media, Coordinator, Task, New Worker).
- **Electron**: Skills scan, write, delete, config management; seed example skills.
- **Frontend**: Skills page (Your skills / Example skills), upload, delete, scope management. Example skills cannot be deleted.
- **Example skills**: code-reviewer, data-analyzer, report-writer.

#### Agents Tab

- New **Agents** tab in the main nav (second position after Projects).
- Agents page with Models, Skills, and Memory sub-tabs.

#### Other

- `agentMap` moved to `agents.tsx`; i18n updates; Models page refactored.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/components/ChatBox/BottomBox/InputBox.tsx
+++ b/src/components/ChatBox/BottomBox/InputBox.tsx
@@ -310,6 +310,7 @@ export const Inputbox = ({
               'border-none shadow-none focus-visible:outline-none focus-visible:ring-0',
               'max-h-[200px] min-h-[40px] px-0 py-0',
               'scrollbar overflow-auto',
+              'placeholder:text-text-label',
               isActive ? 'text-input-text-focus' : 'text-input-text-default'
             )}
             style={{
@@ -477,8 +478,6 @@ export const Inputbox = ({
               value.trim().length > 0 && 'rotate-[-90deg]'
             )}
           />
-          {/* Inner shadow highlight (from Figma design) */}
-          <div className="pointer-events-none absolute inset-0 shadow-[0px_1px_0px_0px_inset_rgba(255,255,255,0.33)]" />
         </Button>
       </div>
     </div>
--- a/src/style/token.css
+++ b/src/style/token.css
@@ -184,7 +184,7 @@
     --input-border-default: var(--border-secondary);
     --input-border-hover: var(--border-action-hover);
     --input-border-focus: var(--border-information);
-    --input-text-default: var(--text-primary);
+    --input-text-default: var(--text-secondary);
     --input-text-focus: var(--text-primary);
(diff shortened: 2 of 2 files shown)


backend/app/service/chat_service.py
```diff
@@ -2183,7 +2190,13 @@ def _create_coordinator_and_task_agents() -> list[ListenChatAgent]:
                                 working_directory=working_directory,
                             )
                         )
-                    ).get_tools()
+                    ).get_tools(),
+                    *SkillToolkit(
+                        options.project_id,
+                        key,
+                        working_directory=working_directory,
+                        user_id=options.user_id,
```
Wendong-Fan: options.user_id may not match the folder key used by the Electron skills config, which is derived from the email prefix. If they differ, enabled state and agent scope may appear saved in UI but not apply at runtime

electron/main/index.ts
```diff
@@ -861,6 +862,378 @@ function registerIpcHandlers() {
...
+      return { success: true };
+    } catch (error: any) {
+      log.error('skill-delete failed', error);
+      return { success: false, error: error?.message };
+    }
+  });
+
+  ipcMain.handle('skill-read', async (_event, filePath: string) => {
+    try {
+      const fullPath = path.isAbsolute(filePath)
```
Wendong-Fan, line 988: Can we limit this read target to SKILLS_ROOT. Accepting an absolute path from renderer can read files outside the skills folder. The same guard would also help skill write, delete, and list.

electron/main/index.ts
```diff
@@ -1466,6 +1840,269 @@ const ensureEigentDirectories = () => {
...
+    if (!existsSync(SKILLS_ROOT)) {
+      await fsp.mkdir(SKILLS_ROOT, { recursive: true });
+    }
+
+    // Step 1: Extract zip into temp directory
+    await fsp.mkdir(tempDir, { recursive: true });
+    const directory = await unzipper.Open.file(zipPath);
+    for (const file of directory.files as any[]) {
+      if (file.type === 'Directory') continue;
+      const destPath = path.join(tempDir, file.path);
```
Wendong-Fan: better add a path boundary check before writing this entry? A crafted zip entry path can include parent traversal segments and escape tempDir, so validating that the resolved destination stays inside tempDir would close this gap.

PR #1236, opened 2026-02-11
fix: allow js scipts for most https origin

### LINKED ISSUE #1233: [Feature Request] Support rendering for HTML 3D assets
### Motivation

We need to add support for rendering HTML-based 3D assets within the desktop.

1. Enable rendering for .html files containing 3D content.
2. Ensure support for interactive elements (rotation, zoom).

https://github.com/user-attachments/assets/e59b11c3-8e7e-4ddb-b34e-9bf487a7ab0a

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

bytecii:
### Description

Sometimes eigent generates the interactive htmls but cannot render / interact usually because of we now only allow one javascript script (js from https://cdn.amplitude.com).

Now add as many as https allowlist as possible

Fixes #1233

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/components/Folder/index.tsx
+++ b/src/components/Folder/index.tsx
@@ -71,7 +71,7 @@ interface FileTreeProps {
   isShowSourceCode: boolean;
 }
 
-const FileTree: React.FC<FileTreeProps> = ({
+export const FileTree: React.FC<FileTreeProps> = ({
   node,
   level = 0,
   selectedFile,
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -6,7 +6,7 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
 		<meta
 			http-equiv="Content-Security-Policy"
-			content="script-src 'self' 'unsafe-inline' 'unsafe-eval' https://cdn.amplitude.com;  worker-src 'self' blob:; child-src 'self' blob:;frame-src 'self' localfile: blob: data:;"
+			content="script-src 'self' 'unsafe-inline' 'unsafe-eval' https:;  worker-src 'self' blob:; child-src 'self' blob:;frame-src 'self' localfile: blob: data:;"
```
nitpicker55555: I believe `https:` is too broad, as it allows scripts from any HTTPS site to be loaded and executed in the application. If the LLM generates a malicious tag, the CSP would not block it. （That is really possible, check https://cybernews.com/security/vibe-coding-dangers-ai-pull-contaminated-content）

Perhaps we could instead introduce a whitelist of commonly used CDNs, such as:

* [https://cdn.amplitude.com/](https://cdn.amplitude.com)
* [https://cdnjs.cloudflare.com/](https://cdnjs.cloudflare.com)
* [https://cdn.jsdelivr.net/](https://cdn.jsdelivr.net)
* [https://unpkg.com/](https://unpkg.com)
* [https://cdn.plot.ly/](https://cdn.plot.ly)
* [https://d3js.org/](https://d3js.org)

bytecii: @nitpicker55555 ok let's for now add as many cds as possible.

nitpicker55555: Thanks for update, but I think we need to remove https://raw.githubusercontent.com, since everyone can upload evil js in github, check https://github.com/rgrove/rawgit/issues/197

bytecii: ok

src/components/Folder/index.tsx
```diff
@@ -1034,11 +1034,13 @@ function HtmlRenderer({
...
           }}
         >
+          {/* allow-same-origin: lets agent-generated HTML access localStorage/sessionStorage (e.g. saving game state).
+               Security is maintained via CSP allowlist in index.html which restricts script sources. */}
           <iframe
             ref={iframeRef}
             srcDoc={processedHtml}
             className="bg-white h-full w-full border-0"
-            sandbox="allow-scripts allow-forms"
+            sandbox="allow-scripts allow-forms allow-same-origin"
```
nitpicker55555: can we remove allow-same-origin? When an iframe enables both allow-scripts and allow-same-origin, and uses srcDoc (thus sharing the same origin as the parent page), scripts inside the frame effectively gain the same privileges as the main application — allowing them to access the parent DOM, read sensitive data, and even remove the sandbox, rendering the isolation meaningless from a security perspective. check https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox

bytecii: If we removed that, many demo examples will not be correctly shown.

Wendong-Fan: would using blob URL instead of srcDoc could resolve this issue?

Wendong-Fan: For this trade-off, the most efficient approach seems to be simply removing allow-same-origin. It’s mainly used for localStorage, which isn’t typically required for 3D rendering. What do you think?
@bytecii

PR #1247, opened 2026-02-12
fix: defer inline scripts in HTML preview so Chart.js and CDN scripts load before chart code runs (#1228)

### LINKED ISSUE #1228: [Feature Request] Bug: Images fail to render within HTML content
### Motivation

Images embedded in HTML components are failing to load or render correctly.

Image

Image

previous html files generated:

[in_progress_priority_impact_report.html](https://github.com/user-attachments/files/25240794/in_progress_priority_impact_report.html)

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

spider-yamet:
## Summary
Fixes charts and script-driven visuals not rendering when opening HTML reports (e.g. `in_progress_priority_impact_report.html`) in the file viewer. Inline scripts were running before external scripts (e.g. Chart.js from CDN), so `Chart` was undefined and canvas charts did not draw.

## Changes
- **`src/lib/htmlFontStyles.ts`**: Added `deferInlineScriptsUntilLoad(html)`. When the document contains at least one `<script src="...">`, it wraps inline `<script>` content in `window.addEventListener('load', ...)` so it runs after all resources (including CDN scripts) have loaded.
- **`src/components/Folder/index.tsx`**: Uses `deferInlineScriptsUntilLoad` when building HTML for the iframe before setting `srcdoc`, so chart-initialization code runs only after Chart.js is available.

## Testing
- Open an HTML report that uses Chart.js (or similar CDN script) with inline chart code in the Folder/file viewer. Charts should render in the preview.
- HTML without external scripts is unchanged (no wrapping).

Closes #1228

### DIFF
--- a/src/lib/htmlFontStyles.ts
+++ b/src/lib/htmlFontStyles.ts
@@ -59,21 +59,61 @@ export function isHtmlDocument(text: string): boolean {
  */
 function isClassicInlineJs(attrs: string): boolean {
   const a = attrs.toLowerCase();
-  if (/type\s*=\s*["']?\s*module\s*["']?/.test(a)) return false;
-  if (/type\s*=\s*["']?\s*application\/ld\+json/.test(a)) return false;
-  return true;
+  const typeMatch = a.match(
+    /(?:^|\s)type\s*=\s*(?:"([^"]*)"|'([^']*)'|([^\s>]+))/i
+  );
+  if (!typeMatch) return true;
+
+  const rawType = (typeMatch[1] ?? typeMatch[2] ?? typeMatch[3] ?? '').trim();
+  if (!rawType) return true;
+
+  const normalizedType = rawType.split(';', 1)[0].trim();
+  if (!normalizedType) return true;
+
+  if (normalizedType === 'module') return false;
+  if (normalizedType === 'application/ld+json') return false;
+
+  const jsMimeTypes = new Set([
+    'text/javascript',
+    'application/javascript',
+    'text/ecmascript',
+    'application/ecmascript',
+    'application/x-javascript',
+    'text/x-javascript',
+    'application/x-ecmascript',
+    'text/x-ecmascript',
+    'text/jscript',
+    'text/livescript',
+  ]);
+
+  return jsMimeTypes.has(normalizedType);
 }
 
-/** Defers inline classic-JS until DOMContentLoaded when external scripts exist; leaves type="module" and application/ld+json unchanged and preserves attributes. */
+/**
+ * Returns true when a script tag has a real src attribute.
+ * This intentionally excludes attributes like data-src.
+ */
+function hasScriptSrc(attrs: string): boolean {
+  return /(?:^|\s)src\s*=/.test(attrs.toLowerCase());
+}
+
+/**
+ * Defers inline classic-JS that appears after external scripts until window load.
+ * This keeps pre-library config scripts in place and preserves global scope by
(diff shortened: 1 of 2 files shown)


spider-yamet:
@Wendong-Fan @bytecii @4pmtong @Pakchoioioi could you please review the PR?

Best Regards

src/lib/htmlFontStyles.ts
```diff
@@ -53,3 +53,55 @@ export function isHtmlDocument(text: string): boolean {
...
+  const lower = html.toLowerCase();
+  let idx = lower.indexOf('<script');
+  let hasExternal = false;
+  while (idx !== -1) {
+    const end = html.indexOf('>', idx);
+    if (end !== -1 && /\bsrc\s*=/.test(html.slice(idx, end))) {
+      hasExternal = true;
+      break;
+    }
+    idx = lower.indexOf('<script', idx + 1);
```
bytecii, line 118: Not all script tags are js scripts right? for example, type="module" and type="application/ld+json"

src/lib/htmlFontStyles.ts
bytecii, line 1: Seems that some script attributes will be removed of the script such as crossorigin etc?

src/lib/htmlFontStyles.ts
```diff
@@ -53,3 +53,55 @@ export function isHtmlDocument(text: string): boolean {
...
+    const contentEnd = html.toLowerCase().indexOf(endTag, contentStart);
+    if (contentEnd === -1) {
+      result += html.slice(scriptStart);
+      break;
+    }
+    const fullTag = html.slice(scriptStart, contentEnd + endTag.length);
+    const content = html.slice(contentStart, contentEnd);
+    if (!hasSrc && content.trim().length > 0) {
+      const escaped = content.replace(/<\/script>/gi, '<\\/script>');
+      result += `<script>window.addEventListener('load',function(){${escaped}});</script>`;
```
bytecii: Should we use DOMContentLoaded instead of load? Instead of waiting all things such as images?

spider-yamet:
@bytecii Let me summarize the update.

**1. Not all script tags are JS**
Added isClassicInlineJs(attrs) so we only wrap classic inline JS.
Scripts with type="module" or type="application/ld+json" are left as-is (no wrapping).
Only classic scripts (no type or text/javascript) are deferred.
**2. Preserve script attributes**
Wrapped output now keeps the original opening tag: we use openTag (from <script through >) and emit ${openTag}window.addEventListener(...)</script>.
Attributes like crossorigin, type="text/javascript", etc. are preserved on deferred scripts.
**3. DOMContentLoaded instead of load**
Replaced window.addEventListener('load', ...) with window.addEventListener('DOMContentLoaded', ...) so we don’t wait for images/resources, only for the DOM and synchronous scripts (e.g. CDN script) to be ready.

spider-yamet:
@Wendong-Fan @bytecii Could you please check PR status?

Wendong-Fan (approved):
thanks for the  fix, while testing, I ran into a few edge cases in the deferral logic: it was delaying inline scripts that should run before external libs, and wrapping inline code in an event callback changed global-scope behavior. I also noticed `src` detection could misfire on things like `data-src`.

  I made a small follow-up that keeps the intent but avoids those regressions:
  - only defer classic inline JS that appears after an external script
  - preserve global execution semantics (dynamic script injection on load)
  - tighten `type/src` checks (including MIME types with parameters)

PR #1251, opened 2026-02-12
fix: resolve false "plan limit reached" error for cloud models

ghost:
## Summary
- Fix incorrect `model_platform` mapping for cloud models routed through LiteLLM proxy — non-GPT models (Gemini, Claude, Minimax) were using native provider SDKs (`gemini`, `anthropic`) instead of `openai-compatible-model`, causing protocol mismatches that were misinterpreted as budget errors
- Remove overly broad `" 429"` heuristic in `error_format.py` that falsely classified HTTP 429 rate limits as `insufficient_quota` errors, and add a separate `rate_limit_exceeded` handler
- Add missing cloud models (GPT-5/5.1/5.2/5-mini, Gemini 3 Flash, Claude Sonnet 4-5) to backend `ModelType` enum

## Test plan
- [ ] Select Minimax M2.5 as cloud model and send a message — should no longer show "plan limit reached"
- [ ] Select Gemini 3 Pro as cloud model and send a message — should work without false quota error
- [ ] Select GPT 5.2 as cloud model and send a message — should work as before
- [ ] Select GPT 4.1 as cloud model and send a message — verify no regression
- [ ] Trigger an actual rate limit (429) and verify the new "Rate limit exceeded" message appears instead of "exceeded your quota"

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -12,7 +12,7 @@ body:
     id: version
     attributes:
       label: What version of eigent are you using?
-      placeholder: E.g., 0.0.83
+      placeholder: E.g., 0.0.84
     validations:
       required: true
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -277,18 +277,74 @@ jobs:
           fi
 
       # Create GitHub Release
+      - name: Prepare GitHub Release assets
+        if: startsWith(github.ref, 'refs/tags/')
+        shell: bash
+        run: |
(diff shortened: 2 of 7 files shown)


ghost:
@Wendong-Fan.
plz help me with this issue

Wendong-Fan (commented):
Thanks @Dang-Ver for the fix! I think a better solution would be to remove the error-formatting module from the original code. In most cases, the API already provides a clear and informative error message. Adding this extra formatting layer introduces additional maintenance overhead and potential bugs.

It would be great if you’re interested in refactoring the current implementation to simply preserve and return the original error message from the API response.

PR #1278, opened 2026-02-16
Fix skill upload bug

Douglasymlai:
### Description

Fix zip and skill file upload error issue

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


electron/main/index.ts
```diff
@@ -1846,20 +1853,111 @@ async function importSkillsFromZip(zipPath: string): Promise<{
...
+          withFileTypes: true,
+        });
+        for (const entry of rootEntries) {
+          if (entry.isFile()) {
+            await fsp.copyFile(
+              path.join(tempDir, entry.name),
+              path.join(dest, entry.name)
+            );
+          }
+        }
```
4pmtong: Currently only root-level files are copied when SKILL.md is at the zip root; subdirectories (e.g. assets/) are not copied and will be lost.
Consider recursively copying all root-level entries (files and subdirectories), similar to the copyDir logic used for the subdirectory branch

electron/main/index.ts
```diff
@@ -1846,20 +1853,111 @@ async function importSkillsFromZip(zipPath: string): Promise<{
...
+              await fsp.mkdir(dstPath, { recursive: true });
+              await copyDir(srcPath, dstPath);
+            } else {
+              await fsp.copyFile(srcPath, dstPath);
+            }
+          }
+        }
+        await copyDir(skillDir, dest);
+      }
+    }
```
4pmtong: When multiple skills in the same zip share the same directory name (e.g. skill-a/SKILL.md and other/skill-a/SKILL.md), both will be written to SKILLS_ROOT/skill-a and the latter will overwrite the former.
Consider adding a suffix when the destination already exists.

4pmtong (changes requested):

PR #1280, opened 2026-02-16
fix(ui): prevent install screen flash when already on main content

dataCenter430:
Closes: #1260 
### Related Issue
Layout visibility race: setWaitingBackend() could briefly show the install
screen after main content had already rendered.

- useInstallationSetup: only call setWaitingBackend() when initState !== 'done'
- Layout: derive install-screen visibility from shouldShowInstallScreen and
  initState only; stop using installationState === 'waiting-backend'.

### Description
- useInstallationSetup: only call setWaitingBackend() when initState !== 'done'
  so we never switch to the install UI if the user is already "done". Still run
  backend polling in the background.
- Layout: derive install-screen visibility from shouldShowInstallScreen and
  initState only; stop using installationState === 'waiting-backend' so that
  a late waiting-backend state cannot override and show the install screen when
  initState is already 'done'.

### Expected behavior
Before: User sees main app (initState is "done"). An async path (e.g. initial “tools installed, wait for backend” or login-after-logout) calls setWaitingBackend(). Layout sees installationState === 'waiting-backend' and shows the install screen, causing a brief flash or full switch back to the install screen.
After: If the user is already "done", we never call setWaitingBackend() for that flow, and Layout no longer uses waiting-backend alone to show the install screen. So the install screen does not appear after main content has rendered. Backend polling still runs; when the backend is ready we stay on main content (or show an error dialog if it fails). During normal first-time setup (initState carousel/permissions), the install screen still shows and “waiting for backend” still appears as before.

### Why it’s better
- Better UX: No more install screen flashing over the main app once the user is already in.
- Correct source of truth: What the user sees is driven by initState (and explicit install flags), not by a late waiting-backend state that can fire after we’re already “done”.
- Defense in depth: Both the hook (don’t set waiting-backend UI when done) and Layout (don’t show install screen when done) are aligned, so the race is fixed and future similar bugs are less likely.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


src/components/Layout/index.tsx
```diff
@@ -73,10 +73,10 @@ const Layout = () => {
...
     initState === 'done' && isFirstLaunch && !isInstalling;
 
+  // Don't show install screen solely due to waiting-backend when already "done"
+  // (avoids race where setWaitingBackend() runs after content has rendered)
   const actualShouldShowInstallScreen =
-    shouldShowInstallScreen ||
-    initState !== 'done' ||
-    installationState === 'waiting-backend';
+    shouldShowInstallScreen || initState !== 'done';
   const shouldShowMainContent = !actualShouldShowInstallScreen;
```
nitpicker55555, line 80: @bytecii @dataCenter430 I would like to understand why the waiting-backend check was removed here. 

This means: as long as the user has previously completed the setup (i.e., `initState === 'done'` has been persisted), the application proceeds directly to the main UI, regardless of whether the backend has started or is ready.

This can result in chatStore being null and consequently no agent workforce being displayed on the frontend (this bug has already been reported by community members). In addition, sending requests while the backend is not yet ready may also lead to runtime errors.

Furthermore, the issue linked in this PR appears to be largely unrelated to the actual problem this PR is addressing.

nitpicker55555:
One fix pr #1348 which fix flash problem and keep waiting-backend

In the old implementation, the race condition occurred because `installationState === 'waiting-backend'` depended on the asynchronous invocation of `setWaitingBackend()`. During the first render frame, the state was still `'idle'`, so the guard did not block execution.

By contrast, `!isBackendReady` evaluates to `true` from the very first frame (since its initial value is `false`), so this approach does not introduce a race condition.

PR #1281, opened 2026-02-16
feat: resolve Zustand persist hydration causing redirect/layout flicker

### LINKED ISSUE #1260: [Question] [Feature Request] Architecture and UX Improvement

dataCenter430:
### Related Issue
Persisted auth may hydrate after the first render, causing temporary redirects or flicker.

Closes #1260 

### Description
1. src/hooks/useAuthHydration.ts (new)
- Hook that returns true only after the auth store’s persist middleware has finished rehydrating.
- Uses useAuthStore.persist.hasHydrated() and onFinishHydration() so we don’t read token / initState / isFirstLaunch before they’re restored from storage.
2. src/routers/index.tsx
- ProtectedRoute uses useAuthHydration().
- Shows the loading spinner until hasHydrated is true, then runs the existing proxy check and INITIALIZE with the real token.
- Prevents briefly sending logged-in users to /login before hydration completes.
3. src/components/Layout/index.tsx
- Layout uses useAuthHydration().
- Keeps showing “Loading…” until the auth store has hydrated, then reads initState and isFirstLaunch.
- Prevents the install/onboarding screen from flashing before switching to the correct main content.

### Why it’s better
- No auth flicker: Route guard and layout only use auth state after rehydration, so we don’t redirect to login or show wrong UI and then correct it.
- Single source of truth: Hydration is handled in one hook (useAuthHydration) and reused wherever auth-dependent UI is rendered.
- Uses Zustand’s API: Relies on persist.hasHydrated() and onFinishHydration() instead of ad-hoc timing or timeouts.
- Same loading UX: We still show a spinner/“Loading…” until we know the real auth state; we only avoid the incorrect redirect and layout flicker that happened when acting on unhydrated state.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


Wendong-Fan:
thanks @dataCenter430 's PR! could @4pmtong and @nitpicker55555 help reviewing this?

nitpicker55555 (changes requested):
**Design concern — solving a likely non-existent problem.**

The auth store uses Zustand `persist` with `localStorage`, which hydrates synchronously. In Electron's renderer process, `hasHydrated()` returns `true` immediately on first render — the "flicker" this PR aims to fix almost certainly doesn't manifest.

Please provide a screen recording showing the actual flicker before/after. If the issue can't be reproduced, this PR adds unnecessary complexity (new hook file, extra loading gates in two components) for a theoretical problem.

Minor: `if (!chatStore) console.log(chatStore)` — logging `null` is meaningless, should be removed.

dataCenter430:
@nitpicker55555  thank you for your review.
I re-tested this before opening the PR.
While localStorage itself is synchronous, I was able to reproduce the flicker occasionally on cold start / refresh.
What I observed:
- First render sees default auth state (token = null / initState default)
- Shortly after, persist rehydrates and updates the store
- ProtectedRoute/Layout briefly acts on the default state before the restored state is applied

It’s not 100% reproducible on every refresh, which suggests it’s a timing/order issue rather than storage latency. But it does occur.
The useAuthHydration() hook simply gates auth-dependent logic on persist.hasHydrated() so we never act on the default state before rehydration completes.
Even if hydration is usually synchronous, this ensures deterministic behavior and removes that race condition entirely.

nitpicker55555:
> @nitpicker55555 thank you for your review. I re-tested this before opening the PR. While localStorage itself is synchronous, I was able to reproduce the flicker occasionally on cold start / refresh. What I observed:
> 
> * First render sees default auth state (token = null / initState default)
> * Shortly after, persist rehydrates and updates the store
> * ProtectedRoute/Layout briefly acts on the default state before the restored state is applied
> 
> It’s not 100% reproducible on every refresh, which suggests it’s a timing/order issue rather than storage latency. But it does occur. The useAuthHydration() hook simply gates auth-dependent logic on persist.hasHydrated() so we never act on the default state before rehydration completes. Even if hydration is usually synchronous, this ensures deterministic behavior and removes that race condition entirely.

However, this is overly complex: creating a separate hook file plus 43 lines of code just to wrap `hasHydrated()` and `onFinishHydration()`, and it’s only used in two places.

A simpler approach would be to add a `_hasHydrated` state field directly to `authStore`, and set it to `true` inside the `onRehydrateStorage` callback of `persist`. Consumers can then simply use `useAuthStore(s => s._hasHydrated)`.

No new file, no additional hook required.

4pmtong:
Hi @dataCenter430, thanks for the contribution and the thorough write-up! We've reviewed this carefully, and our concern is that the change adds a new hook and
  hydration guards in multiple components to solve a problem that's largely theoretical in our case — since we use localStorage (synchronous), Zustand's persist
  middleware hydrates before the first render, so the redirect/flicker shouldn't happen in practice.

  We'd prefer to keep the codebase lean and only add these guards if we have a reproducible case. If you've been able to consistently reproduce the flicker, could
  you share the steps or a screen recording? That would help us reconsider.

  Also a note on #1260 — since that issue covers 6 different items, it might be better to open separate issues for each one so they can be discussed and reviewed
  independently. Thanks again!

PR #1283, opened 2026-02-16
fix(backend): restrict CORS to configurable origins and explicit methods/headers

### LINKED ISSUE #1260: [Question] [Feature Request] Architecture and UX Improvement

dataCenter430:
### Related Issue
CORS configuration: missing or overly permissive CORS can allow unwanted origins.

Closes #1260

### Description
Replace permissive CORS (allow_origins=["*"] with allow_credentials=True and
allow_methods/allow_headers=["*"]) with a safe, configurable setup:

- Origins: read from CORS_ORIGINS (comma-separated) in ~/.eigent/.env or env.
  In development, if unset, allow only http://localhost:5173, 127.0.0.1:5173,
  and ports 3000. In non-development, allow no origins until CORS_ORIGINS is set.
- Methods: GET, POST, PUT, DELETE, OPTIONS only.
- Headers: Content-Type, Authorization, x-stack-auth only.

Document CORS_ORIGINS in backend README. Ensures credentials work correctly
(no wildcard origin with credentials) and reduces risk of unwanted origins
calling the API from the browser.

### Why?
- Security: Overly permissive CORS lets any website send credentialed requests to your API from the browser; restricting to an explicit origin list and to the methods/headers the app uses reduces that risk.
- Correctness: Using allow_origins=["*"] with allow_credentials=True is invalid per the CORS spec; browsers may ignore or mishandle it. Using a concrete list of origins with credentials is valid and predictable.
- Operability: Development keeps working with default localhost origins; production stays strict until CORS_ORIGINS is set, and the README explains how to configure it.

- [x] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -2,6 +2,12 @@
 uv run uvicorn main:api --port 5001
 ```
 
+### CORS
+
+CORS is restricted to avoid allowing arbitrary origins. Configure allowed browser origins in `~/.eigent/.env`:
+
+- **`CORS_ORIGINS`** (optional): Comma-separated list of origins, e.g. `http://localhost:5173,http://localhost:3000`. In development, if unset, common localhost origins are allowed; in production, no origins are allowed until you set this.
+
 i18n operation process: https://github.com/Anbarryprojects/fastapi-babel
 
 ```bash
--- a/backend/app/__init__.py
+++ b/backend/app/__init__.py
@@ -12,17 +12,55 @@
 # limitations under the License.
 # ========= Copyright 2025-2026 @ Eigent.ai All Rights Reserved. =========
 
+import logging
+import os
+
+from dotenv import load_dotenv
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
+
+# Load env so CORS_ORIGINS is available when this module is imported (before main runs).
+load_dotenv(dotenv_path=os.path.join(os.path.expanduser("~"), ".eigent", ".env"))
+
+logger = logging.getLogger(__name__)
 
 # Initialize FastAPI with title
 api = FastAPI(title="Eigent Multi-Agent System API")
 
-# Add CORS middleware
-api.add_middleware(
-    CORSMiddleware,
-    allow_origins=["*"],
-    allow_credentials=True,
-    allow_methods=["*"],
-    allow_headers=["*"],
-)
+
+def _get_cors_origins() -> list[str]:
+    """
+    CORS allowed origins. Avoids overly permissive '*' to prevent unwanted origins.
+    Set CORS_ORIGINS (comma-separated) in ~/.eigent/.env or environment;
+    in development only, defaults to common localhost origins if unset.
+    """
+    raw = os.environ.get("CORS_ORIGINS")
+    if raw is not None and raw.strip():
+        origins = [o.strip() for o in raw.split(",") if o.strip()]
+        if origins:
+            return origins
+    if os.environ.get("ENVIRONMENT", "development").lower() == "development":
+        return [
+            "http://localhost:5173",
+            "http://127.0.0.1:5173",
+            "http://localhost:3000",
+            "http://127.0.0.1:3000",
+        ]
+    return []
+
+
+_cors_origins = _get_cors_origins()
+if _cors_origins:
+    api.add_middleware(
+        CORSMiddleware,
+        allow_origins=_cors_origins,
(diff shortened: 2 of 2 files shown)


Wendong-Fan:
thanks @dataCenter430 for the PR, could @4pmtong and @nitpicker55555 help reviewing this?

dataCenter430:
Thanks for the feedback. @nitpicker55555 
This PR isn’t meant to address a remote attacker threat in the current Electron-only, localhost-bound deployment. The intent is defense-in-depth and future-proofing in case the backend is:
• Reused outside Electron
• Exposed via containerization or dev tooling
• Accessed from a browser context (intentionally or accidentally)
CORS here acts as a guardrail, not a hard security boundary. The env-var configuration allows behavior to be adjusted without code changes.
Regarding allow_headers: the explicit list is intentional to keep the API surface strict. If the team prefers flexibility, I’m happy to relax this to *.

If we’re confident the backend will never be used outside the current model, I’m okay but I wanted to add a minimal safety net for likely future scenarios.
@Wendong-Fan @nitpicker55555  Therefore, I want best practicies. 
Best regards

nitpicker55555:
> Thanks for the feedback. @nitpicker55555 This PR isn’t meant to address a remote attacker threat in the current Electron-only, localhost-bound deployment. The intent is defense-in-depth and future-proofing in case the backend is: • Reused outside Electron • Exposed via containerization or dev tooling • Accessed from a browser context (intentionally or accidentally) CORS here acts as a guardrail, not a hard security boundary. The env-var configuration allows behavior to be adjusted without code changes. Regarding allow_headers: the explicit list is intentional to keep the API surface strict. If the team prefers flexibility, I’m happy to relax this to *.
> 
> If we’re confident the backend will never be used outside the current model, I’m okay but I wanted to add a minimal safety net for likely future scenarios. @Wendong-Fan @nitpicker55555 Therefore, I want best practicies. Best regards

"defense-in-depth" argument doesn't apply here — defense-in-depth requires an actual threat surface, and a localhost-only API has none. Future-proofing for hypothetical scenarios (containerization, external exposure) is YAGNI — if the architecture changes, CORS requirements will be different anyway, and can be added at that point with the real constraints in mind. Best practice is matching security measures to the actual threat model, not applying web patterns to desktop apps. Suggest to close.

dataCenter430:
Thanks for the thoughtful feedback, I agree that security controls should match the actual threat model, and that a localhost-only API does not present a traditional remote attack surface.
To clarify, my intent wasn’t to position this as protection against external attackers or to apply web security patterns blindly. The change is mainly about explicitly defining the expected browser origins for the local API and avoiding implicit assumptions about runtime context (dev server vs packaged Electron).
Even though the backend binds to localhost, the renderer still runs in a browser context, and CORS continues to define which browser origins are allowed to communicate with the API. In practice, this serves more as explicit contract and isolation of intended usage rather than as a primary security boundary.
That said, I understand the YAGNI concern. If the team prefers to keep the backend completely minimal given the current architecture, I’m fine closing this and reintroducing CORS only if/when the exposure model changes.
Happy to align with the team’s direction here, thx again @nitpicker55555

nitpicker55555:
> Thanks for the thoughtful feedback, I agree that security controls should match the actual threat model, and that a localhost-only API does not present a traditional remote attack surface. To clarify, my intent wasn’t to position this as protection against external attackers or to apply web security patterns blindly. The change is mainly about explicitly defining the expected browser origins for the local API and avoiding implicit assumptions about runtime context (dev server vs packaged Electron). Even though the backend binds to localhost, the renderer still runs in a browser context, and CORS continues to define which browser origins are allowed to communicate with the API. In practice, this serves more as explicit contract and isolation of intended usage rather than as a primary security boundary. That said, I understand the YAGNI concern. If the team prefers to keep the backend completely minimal given the current architecture, I’m fine closing this and reintroducing CORS only if/when the exposure model changes. Happy to align with the team’s direction here, thx again @nitpicker55555

Electron’s renderer accesses the backend via `http://localhost:5173` in development or `file:///app://` in production, and the current `allow_origins=["*"]` already covers all these cases.

The “problem” you’re trying to solve — restricting which origins can access the backend — does not present a realistically exploitable attack surface in a desktop application context.

PR #1290, opened 2026-02-17
fix: production safety audit — 4 critical bugs across backend, server, and electron

### LINKED ISSUE #1327: [BUG] Production safety: context isolation bypass, JWT error codes, assert control flow, stale background tasks
### Summary

Four confirmed production bugs found across the Python backend, server auth, and Electron main process during a safety audit.

#### 1. CRITICAL - Electron Child Windows Missing Context Isolation (electron/main/index.ts)
The main window correctly sets `contextIsolation: true`, but the `open-win` IPC handler creates child `BrowserWindow` instances with `nodeIntegration: true` and `contextIsolation: false`. Any XSS payload or injected script in a child renderer window gains full `require('child_process').exec(...)` access with no sandbox barrier.

**Impact**: Full remote code execution from any child window.

#### 2. HIGH - JWT Token Expiry Returns Wrong Error Code (server/app/component/auth.py)
PyJWT's `jwt.decode()` raises `ExpiredSignatureError` (a subclass of `InvalidTokenError`) for expired tokens. The manual `payload["exp"] < datetime.now()` check is dead code. The `except InvalidTokenError` block catches both expired and invalid tokens, returning `token_invalid` for both. Clients cannot distinguish "token expired, please refresh" from "token is corrupted, please re-login".

**Impact**: Broken automatic token refresh logic for all clients.

#### 3. HIGH - assert Statements Used for Runtime Control Flow (backend/app/agent/listen_chat_agent.py)
`step()` and `astep()` use `assert message is not None` and `assert res is not None` for critical runtime checks. Python's `assert` statements are completely removed when running with `-O` (optimized mode), common in production. When stripped, `None` message corrupts the SSE data stream and `None` response causes downstream `AttributeError` crashes.

**Impact**: Silent data corruption and crashes in production with `-O` flag.

#### 4. HIGH - background_tasks.clear() Doesn't Cancel Running Tasks (backend/app/controller/chat_controller.py)
When a user sends a follow-up message after task completion, `improve()` calls `task_lock.background_tasks.clear()` which only empties the set reference without cancelling the actual running coroutines. Stale tasks continue executing in the background.

**Impact**: Resource waste and potential data corruption from stale tasks.

### Affected Files
- `electron/main/index.ts` (1 bug)
- `server/app/component/auth.py` (1 bug)
- `backend/app/agent/listen_chat_agent.py` (1 bug)
- `backend/app/controller/chat_controller.py` (1 bug)

eren-karakus0:
### Related Issue

Closes #1327

## Summary

A cross-stack production safety audit uncovering 4 confirmed bugs across the Python backend, server auth, and Electron main process. Each bug is independently verified and includes test coverage where applicable.

### Bug 1: Electron Child Windows Missing Context Isolation (CRITICAL)

**File:** `electron/main/index.ts` — `open-win` IPC handler

The main window correctly sets `contextIsolation: true`, but the `open-win` handler creates child `BrowserWindow` instances with `nodeIntegration: true` and `contextIsolation: false`. With context isolation disabled, any JavaScript running in a child renderer window shares the same context as Electron's built-in modules — meaning any XSS payload or injected script gains full `require('child_process').exec(...)` access with no sandbox barrier.

**Before:**
```typescript
webPreferences: {
  preload,
  nodeIntegration: true,      // full Node.js access
  contextIsolation: false,    // no sandbox
}
```

**After:**
```typescript
webPreferences: {
  preload,
  nodeIntegration: false,     // no direct Node.js access
  contextIsolation: true,     // isolated contexts
  sandbox: true,              // OS-level sandboxing
}
```

### Bug 2: JWT Token Expiry Returns Wrong Error Code (HIGH)

**File:** `server/app/component/auth.py` — `Auth.decode_token()`

PyJWT's `jwt.decode()` already validates the `exp` claim and raises `ExpiredSignatureError` (a subclass of `InvalidTokenError`) for expired tokens. The manual `payload["exp"] < datetime.now()` check never executes because expired tokens don't reach that line. The `except InvalidTokenError` block catches both expired and invalid tokens, raising `token_invalid` for both cases.

This makes it impossible for clients to distinguish "token expired, please refresh" from "token is corrupted, please re-login", breaking any automatic token refresh logic.

**Before:**
```python
try:
    payload = jwt.decode(token, Auth.SECRET_KEY, algorithms=["HS256"])
    id = payload["id"]
    if payload["exp"] < int(datetime.now().timestamp()):  # dead code
        raise TokenException(code.token_expired, ...)
except InvalidTokenError:  # catches ExpiredSignatureError too
    raise TokenException(code.token_invalid, ...)  # wrong code for expired
```

**After:**
```python
try:
    payload = jwt.decode(token, Auth.SECRET_KEY, algorithms=["HS256"])
except ExpiredSignatureError:
    raise TokenException(code.token_expired, ...)   # correct code
except InvalidTokenError:
    raise TokenException(code.token_invalid, ...)   # only truly invalid tokens
```

### Bug 3: `assert` Statements Used for Runtime Control Flow (HIGH)

**File:** `backend/app/agent/listen_chat_agent.py` — `step()` and `astep()`

Both `step()` and `astep()` use `assert message is not None` and `assert res is not None` for critical runtime checks. Python's `assert` statements are **completely removed** when running with `-O` (optimized mode), which is common in production deployments.

When stripped:
- `None` message propagates into the SSE `ActionDeactivateAgentData` event, corrupting the frontend data stream
- `None` response returns to callers, causing downstream `AttributeError` crashes

**Before:**
```python
assert message is not None    # stripped with python -O
# ...
assert res is not None         # stripped with python -O
return res
```

**After:**
```python
if message is None:
    message = ""
    logger.warning(f"Agent {self.agent_name}: message is None ...")

# ...
if res is None:
    raise RuntimeError(
        f"Agent {self.agent_name}: step() returned None without setting error_info"
    )
return res
```

### Bug 4: `background_tasks.clear()` Doesn't Cancel Running Tasks (HIGH)

**File:** `backend/app/controller/chat_controller.py` — `improve()` handler

When a user sends a follow-up message after task completion, `improve()` calls `task_lock.background_tasks.clear()`. This only removes references from the Python set — it does **not** cancel the underlying `asyncio.Task` objects. Ghost tasks continue running in the event loop, holding resources, and potentially writing stale SSE events into the new conversation's stream.

**Before:**
```python
if hasattr(task_lock, "background_tasks"):
    task_lock.background_tasks.clear()    # tasks keep running
```

**After:**
```python
if hasattr(task_lock, "background_tasks"):
    for bg_task in list(task_lock.background_tasks):
        if not bg_task.done():
            bg_task.cancel()
    task_lock.background_tasks.clear()
```

## Test Plan

Added 6 new unit tests:

**Assert Guard Tests** (`test_listen_chat_agent.py`):
- [x] `test_step_handles_none_message_without_assert` — verifies step() doesn't crash with AssertionError
- [x] `test_step_raises_runtime_error_when_res_is_none` — verifies RuntimeError instead of AssertionError
- [x] `test_astep_handles_none_message_without_assert` — async version of message guard
- [x] `test_astep_raises_runtime_error_when_res_is_none` — async version of res guard

**Background Task Tests** (`test_chat_controller.py`):
- [x] `test_improve_cancels_background_tasks_when_done` — verifies running tasks are cancelled, done tasks are not
- [x] `test_improve_handles_missing_background_tasks_attr` — verifies graceful handling without the attribute

## Files Changed

| File | Change |
|------|--------|
| `electron/main/index.ts` | Set `nodeIntegration: false`, `contextIsolation: true`, `sandbox: true` in child windows |
| `server/app/component/auth.py` | Catch `ExpiredSignatureError` separately; remove dead manual exp check |
| `backend/app/agent/listen_chat_agent.py` | Replace 4 `assert` statements with `if`/`raise` guards |
| `backend/app/controller/chat_controller.py` | Cancel background tasks before clearing the set |
| `backend/tests/app/agent/test_listen_chat_agent.py` | +4 tests for assert replacement |
| `backend/tests/app/controller/test_chat_controller.py` | +2 tests for task cancellation |

### DIFF
--- a/backend/app/agent/listen_chat_agent.py
+++ b/backend/app/agent/listen_chat_agent.py
@@ -292,7 +292,12 @@ class ListenChatAgent(ChatAgent):
                 f"tokens used: {total_tokens}"
             )
 
-        assert message is not None
+        if message is None:
+            message = ""
+            logger.warning(
+                f"Agent {self.agent_name}: message is None after step "
+                f"completion, defaulting to empty string"
+            )
 
         _schedule_async_task(
             task_lock.put_queue(
@@ -310,7 +315,11 @@ class ListenChatAgent(ChatAgent):
 
         if error_info is not None:
             raise error_info
-        assert res is not None
+        if res is None:
+            raise RuntimeError(
+                f"Agent {self.agent_name}: step() returned None "
+                f"without setting error_info"
+            )
         return res
 
     async def astep(
@@ -397,7 +406,12 @@ class ListenChatAgent(ChatAgent):
 
         # Send deactivation for all non-streaming cases (success or error)
         # Streaming responses handle deactivation in _astream_chunks
-        assert message is not None
+        if message is None:
+            message = ""
+            logger.warning(
+                f"Agent {self.agent_name}: message is None after astep "
+                f"completion, defaulting to empty string"
+            )
 
         asyncio.create_task(
             task_lock.put_queue(
@@ -415,7 +429,11 @@ class ListenChatAgent(ChatAgent):
 
         if error_info is not None:
             raise error_info
-        assert res is not None
+        if res is None:
+            raise RuntimeError(
+                f"Agent {self.agent_name}: astep() returned None "
+                f"without setting error_info"
+            )
         return res
 
     def _execute_tool(
--- a/backend/app/agent/listen_chat_agent.py
+++ b/backend/app/agent/listen_chat_agent.py
@@ -293,10 +293,9 @@ class ListenChatAgent(ChatAgent):
             )
 
         if message is None:
-            message = ""
-            logger.warning(
+            raise RuntimeError(
                 f"Agent {self.agent_name}: message is None after step "
-                f"completion, defaulting to empty string"
+                f"completion — this indicates a missing LLM response"
             )
 
         _schedule_async_task(
@@ -407,10 +406,9 @@ class ListenChatAgent(ChatAgent):
         # Send deactivation for all non-streaming cases (success or error)
         # Streaming responses handle deactivation in _astream_chunks
         if message is None:
-            message = ""
-            logger.warning(
+            raise RuntimeError(
                 f"Agent {self.agent_name}: message is None after astep "
-                f"completion, defaulting to empty string"
+                f"completion — this indicates a missing LLM response"
             )
 
         asyncio.create_task(
--- a/backend/tests/app/agent/test_listen_chat_agent.py
+++ b/backend/tests/app/agent/test_listen_chat_agent.py
@@ -558,9 +558,7 @@ async def test_astep_raises_runtime_error_on_none_message(mock_task_lock):
 
         error = Exception("Async unexpected error")
         with patch.object(ChatAgent, "astep", side_effect=error):
-            with pytest.raises(
-                Exception, match="Async unexpected error"
-            ):
+            with pytest.raises(Exception, match="Async unexpected error"):
                 await agent.astep("test input")
 
 
--- a/backend/tests/app/agent/test_listen_chat_agent.py
+++ b/backend/tests/app/agent/test_listen_chat_agent.py
@@ -481,132 +481,115 @@ class TestListenChatAgent:
 
 
 @pytest.mark.unit
-class TestAssertReplacedWithProperGuards:
-    """Verify that assert statements have been replaced with proper guards.
-
-    Python's assert statements are stripped when running with -O (optimized
-    mode). The original code used `assert message is not None` and
-    `assert res is not None` for runtime control flow, which would silently
-    produce None propagation in production builds.
-    """
-
-    def test_step_handles_none_message_without_assert(self, mock_task_lock):
-        """step() should not crash with AssertionError when message is None.
-
-        When super().step() raises an unexpected exception that doesn't set
-        message, the code should handle it gracefully instead of relying on
-        assert.
-        """
-        with (
-            patch(f"{_LCA}.get_task_lock", return_value=mock_task_lock),
-            patch("camel.models.ModelFactory.create") as mock_create_model,
-        ):
-            mock_backend = MagicMock()
-            mock_backend.model_type = "gpt-4"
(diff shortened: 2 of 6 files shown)


backend/app/agent/listen_chat_agent.py
```diff
@@ -292,7 +292,12 @@ def step(
                 f"tokens used: {total_tokens}"
             )
 
-        assert message is not None
+        if message is None:
```
bytecii, line 295: Why we change to warning and let the step continue?

backend/tests/app/agent/test_listen_chat_agent.py
```diff
@@ -480,6 +480,135 @@ def test_listen_chat_agent_step_with_task_lock_error(self):
                 agent.step("Test message")
 
 
+@pytest.mark.unit
+class TestAssertReplacedWithProperGuards:
```
bytecii: Just use `def test_xxx` instead of creating class for now

bytecii (commented):
Thanks.

backend/app/agent/listen_chat_agent.py
eren-karakus0: Good point — changed to raise RuntimeError instead, keeping fail-fast behavior consistent with the res=None guard below. Fixed in latest push.

backend/tests/app/agent/test_listen_chat_agent.py
eren-karakus0: Done — replaced class with plain def test_xxx functions.

Wendong-Fan:
could @eren-karakus0 link one issue this pr is going to resolve?

eren-karakus0:
Linked - closes #1327. Sorry for the delay!

eren-karakus0:
Hey @bytecii, just a friendly ping - all review feedback has been addressed and CI is green (7/7). Could you take another look when you get a chance? Thanks!

eren-karakus0:
Friendly ping - just checking if there's anything I should update or adjust in this PR. Happy to address any feedback\!

PR #1310, opened 2026-02-19
feat: Replace terminal dangerous-command block with HITL approval and add Safe Mode in Settings

### LINKED ISSUE #1306: [Feature Request] Replace strict command prohibition in Terminal Toolkit with structured Human-in-the-Loop verification
### Motivation

Currently, the Terminal Toolkit hard-blocks certain “dangerous” commands, which causes some tasks to fail completely. We need to remove this strict prohibition and replace it with a Human-in-the-Loop (HITL) approval mechanism.

### Solution

The HITL: When any of the following dangerous commands is detected, the system should return three options for the user to choose from:

**Yes** (approve this command once)

**All Yes in this task** (approve all subsequent dangerous commands within the current task)

**No** (reject the command)

These options should be displayed and handled by the frontend.

very important!!!
Do NOT directly reuse the previous human_toolkit activation mechanism that waits for user input and automatically skips after 30 seconds.

### Alternatives

_No response_

### Additional context

**Dangerous Command List (Triggers HITL):**
- System Administration: sudo, su, reboot, shutdown, halt, poweroff, init
- File System: rm, chown, chgrp, umount, mount
- Disk Operations: dd, mkfs, fdisk, parted, fsck, mkswap, swapon, swapoff
- Process Management: service, systemctl, systemd
- Network Configuration: iptables, ip6tables, ifconfig, route, iptables-save
- Cron/Scheduling: crontab, at, batch
- User/Kernel Management: useradd, userdel, usermod, passwd, chpasswd, newgrp, modprobe, rmmod, insmod, lsmod

**Additional Safety Constraint:**
Non-Docker Mode: If the environment is running outside of Docker, the system must validate cd commands. Ensure the agent cannot traverse outside the designated working_directory.

spider-yamet:
## Related issue

Closes https://github.com/eigent-ai/eigent/issues/1306

## Summary

Replaces the Terminal Toolkit’s hard block of “dangerous” commands with a **Human-in-the-Loop (HITL)** approval flow and adds a **Safe Mode** setting so users can opt in.

### HITL for dangerous commands
- When a dangerous command is detected and Safe Mode is on, the user is offered three choices:
  - **Yes** – approve this command once
  - **All Yes in this task** – approve all subsequent dangerous commands in the current task
  - **No** – reject the command
- Frontend shows the prompt and calls `/chat/{id}/terminal-approval` with the chosen option; backend enforces approval before running the command and supports “approve all in task” per task.

### Safe Mode in Settings
- **Settings → Permissions**: new **Safe Mode** toggle with hint:  
  *“With Safe Mode active, Eigent will pause and seek explicit approval whenever high-risk system operations are detected.”*
- Default: **off**. When on, the HITL flow above is used for the configured dangerous-command list.

### Backend
- **Dangerous command list** (triggers HITL when Safe Mode is on): system (e.g. `sudo`, `su`, `reboot`, `shutdown`), file (e.g. `rm`, `chown`, `mount`), disk (e.g. `dd`, `mkfs`, `fdisk`), process (e.g. `service`, `systemctl`), network (e.g. `iptables`, `ifconfig`), cron (e.g. `crontab`, `at`), user/kernel (e.g. `useradd`, `modprobe`), and related commands as specified.
- **Non-Docker mode**: `cd` is validated so the agent cannot leave the designated `working_directory`.
- Task lock includes `approved_all_dangerous_commands` and a queue for terminal approval; reset on new task.

### Frontend
- Permissions tab and Safe Mode UI; preference stored in `localStorage` and sent as `safe_mode` in the start-task request.

### Other
- Pre-commit: skip backend checks when `uv` is not installed so commits succeed without `uv` (message suggests installing uv for backend checks).

## Testing

- [ ] Verified Safe Mode off: dangerous commands run without approval.
- [ ] Verified Safe Mode on: dangerous commands trigger the three-option prompt; Yes / All Yes in task / No behave as described.
- [ ] Verified Permissions UI: toggle and hint render; state persists and is sent to backend.
- [ ] Verified `cd` outside `working_directory` in non-Docker mode is rejected.


spider-yamet:
@bytecii @Wendong-Fan Could you please review my PR?

bytecii (commented):
IMO we can move this to camel to make a special safe mode such as strict. cc @Wendong-Fan

Wendong-Fan:
> IMO we can move this to camel to make a special safe mode such as strict. cc @Wendong-Fan

thanks @bytecii , the idea behind this issue is to implement a human-in-the-loop mechanism for safer code execution, which would involve the HumanToolkit. Since CAMEL serves more as a modular framework, it might be more appropriate to implement this feature in Eigent as an application layer. What do you think?

spider-yamet:
@Wendong-Fan @bytecii could you please review this pr?
I see this PR remains opening for several days.

Regards

backend/app/agent/factory/browser.py
```diff
@@ -226,7 +226,6 @@ def browser_agent(options: Chat):
         options.project_id,
         Agents.browser_agent,
         working_directory=working_directory,
-        safe_mode=True,
```
bytecii, line 229: remove this as this is not configured. And by default if the HITL not specified, we still use the safe_mode=True for the terminaltoolkit

backend/app/agent/toolkit/terminal_toolkit.py
```diff
@@ -349,7 +367,35 @@ def _run_coro_in_thread(coro, task_lock):
...
+        logger.info("[APPROVAL] Received response: %s", approval)
+
+        if approval == ApprovalAction.approve_once:
+            return None
+        if approval == ApprovalAction.auto_approve:
+            task_lock.auto_approve[self.agent_name] = True
+            return None
+        return "Operation rejected by user. The task is being stopped."
+
+    async def shell_exec(
```
bytecii, line 458: changed to async

backend/app/utils/listen/toolkit_listen.py
```diff
@@ -355,6 +361,37 @@ async def async_wrapper(*args, **kwargs):
...
+            # loop. The metadata (name, signature, docstring) is still
+            # preserved from `wrap` by @wraps; only the unwrap chain changes.
+            #
+            # This only triggers when wrap != func, i.e. when a subclass
+            # provides an async override of a sync base method — a pattern
+            # first introduced for TerminalToolkit.shell_exec to support
+            # the HITL approval flow (Camel's upstream shell_exec is
+            # sync-only with no async variant available).
+            if wrap is not func:
+                async_wrapper.__wrapped__ = func
```
bytecii, line 393: Need someone helps to take a look at this

bytecii:
Update the PR including backend and ui redesign
In general, making the HITL as an option in the settings and make it extensible for future possible approval

image
image

- Added a `_request_user_approval()` method on `AbstractToolkit` that any toolkit can call to pause execution and wait for user approval via SSE.
- Now only support the `TerminalToolkit` uses it to gate dangerous commands (rm, sudo, kill, etc.) behind a frontend toggle in the new HITL settings page, with per-agent approval queues so multi-agent tasks don't block each other.
- The auto approval is defined for each task per agent and each task will fetch the terminal toolkit hitl settings to check whether need to let user input the approval for potential dangerous terminal command

tested with and without terminal approval in the HITL + approval / auto approval / reject with & without dark mode
cc @Wendong-Fan @fengju0213

backend/app/agent/toolkit/terminal_toolkit.py
```diff
@@ -349,7 +366,96 @@ def _run_coro_in_thread(coro, task_lock):
...
+        enabled = (
+            task_lock.hitl_options.terminal_approval if task_lock else False
+        )
+        self.safe_mode = not enabled
+        return enabled
+
+    async def _request_user_approval(self, action_data) -> str | None:
+        """Send a command approval request to the frontend and wait.
+
+        Flow::
```
bytecii, line 386: here is the flow

backend/app/service/task.py
```diff
@@ -407,6 +434,85 @@ async def get_human_input(self, agent: str):
...
+        loop = asyncio.get_running_loop()
+        future = loop.create_future()
+        self.pending_approvals[approval_id] = future
+        logger.debug(
+            "Created approval future",
+            extra={"task_id": self.id, "approval_id": approval_id},
+        )
+        return future
+
+    def resolve_approval(self, approval_id: str, action: str):
```
fengju0213, line 497: Approval callbacks may execute outside the Future owner loop/thread; the current restore paths here have a cross-thread wake-up risk, which can lead to “API succeeds but command execution does not continue”.

backend/app/hitl/terminal_command.py
```diff
@@ -0,0 +1,265 @@
...
+        try:
+            work_real = os.path.realpath(os.path.abspath(working_directory))
+            if os.path.isabs(target):
+                resolved = os.path.realpath(os.path.abspath(target))
+            else:
+                resolved = os.path.realpath(
+                    os.path.abspath(os.path.join(work_real, target))
+                )
+            if os.path.commonpath([resolved, work_real]) != work_real:
+                return (
```
fengju0213, line 269: If relative `cd` in compound commands is not resolved against step-by-step directory state, valid sequences (e.g., `cd sub && cd ..`) can be incorrectly flagged as escaping the workspace.

fengju0213 (commented):
thanks ！@spider-yamet and @bytecii  left some comments below and pushed a commit
overall，looks good

backend/app/service/task.py
spider-yamet: We fixed the cross-thread wake-up risk by changing how approval resolution works:

1. No resolution from the API thread – resolve_approval (and the other resolve entrypoints) no longer call future.set_result(...) directly. They only call _resolve_future_threadsafe.
2. Resolution on the Future’s loop – _resolve_future_threadsafe gets the event loop that owns the Future via future.get_loop(), then schedules a callback on that loop with loop.call_soon_threadsafe(self._set_future_result_if_pending, future, action). The callback runs on the loop where the command is awaiting approval and only then calls future.set_result(action), so the wake-up always happens on the correct loop.
3. Defensive checks and logging – We check future.done() and loop.is_closed(), handle RuntimeError from get_loop() and from call_soon_threadsafe (e.g. if the loop closes between the check and scheduling), and log warnings so “API succeeds but command execution does not continue” is visible if something goes wrong.

backend/app/hitl/terminal_command.py
spider-yamet: We’ve addressed this by validating compound cd commands against a step-by-step directory state:

1. State per sub-command – We iterate over each sub-command from split_compound_command and keep a simulated current_dir and previous_dir (both initially set to the canonical workspace path).
2. Relative cd uses current dir – For each cd sub-command we resolve the target: if it’s relative we resolve it with os.path.join(current_dir, target) and then os.path.realpath(os.path.abspath(...)), so cd sub and cd .. are evaluated in the context of the directory left by the previous step.
3. cd - – We treat cd - as “go to previous directory” by setting the target to previous_dir in the simulated state.
4. State updates – After a valid cd we set previous_dir = current_dir and current_dir = resolved before the next sub-command, so a sequence like cd sub && cd .. is validated step-by-step and correctly stays inside the workspace instead of being flagged as escaping.

spider-yamet:
@fengju0213 @Wendong-Fan @bytecii I updated all the details by following the feedback, ready for review. :)

spider-yamet:
could you please check @fengju0213 ? Hope we can complete this task ASAP...

spider-yamet:
Could you let me know your plan, @fengju0213 ?

spider-yamet:
Still waiting for review, @Wendong-Fan @bytecii.. Regards

spider-yamet:
@fengju0213 Could you please review current PR? or hope you can let me know about any timeline for this, remaining for review status for several weeks.. Regards.

fengju0213:
> @fengju0213 Could you please review current PR? or hope you can let me know about any timeline for this, remaining for review status for several weeks.. Regards.

Sorry for the delay. Our team has been extremely busy over the past few weeks. On the code side, we don't have any major issues at the moment. However, since this involves UI and user experience, we need our product team to review the UI part. I will ask them to provide feedback as soon as possible.

Once again, apologies for the delay!

spider-yamet:
@bytecii Could you please review this PR?

PR #1348, opened 2026-02-23
Fix: chatstore null

nitpicker55555:
### Related Issue

Closes #

### Testing Evidence (REQUIRED)

- [ ] I have included human-verified testing evidence in this PR.
- [ ] This PR includes frontend/UI changes, and I attached screenshot(s) or screen recording(s).
- [ ] No frontend/UI changes in this PR.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

- [ ] I have read and agree to the [Eigent Contribution Guideline](https://github.com/eigent-ai/eigent/blob/main/CONTRIBUTING.md#eigent-contribution-guideline)

### DIFF
--- a/src/store/installationStore.ts
+++ b/src/store/installationStore.ts
@@ -51,7 +51,6 @@ interface InstallationStoreState {
   setError: (error: string) => void;
   setBackendError: (error: string) => void;
   setWaitingBackend: () => void;
-  setBackendReadyFlag: (ready: boolean) => void;
   setNeedsBackendRestart: (needs: boolean) => void;
   retryInstallation: () => void;
   retryBackend: () => Promise<void>;
@@ -90,6 +89,7 @@ export const useInstallationStore = create<InstallationStoreState>()(
         progress: 20,
         logs: [],
         error: undefined,
+        isBackendReady: false,
         isVisible: true,
       }),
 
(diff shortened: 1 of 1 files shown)


nitpicker55555:
PR #1280 introduce a bug, it removes the `waiting-backend` guard. After an application restart, `initState` is restored from `localStorage` as `'done'`. The `Layout` component therefore assumes initialization has completed and immediately renders the main UI.

As part of this rendering process, `useChatStoreAdapter()` is invoked synchronously. At this moment, the backend is not yet ready, but `projectStore.getActiveChatStore()` is already executed. Because the `useEffect` subscription has not been triggered yet, the initial `chatState` remains `null`. As a result, `chatStore` is `null`, and accessing `.tasks` throws a `TypeError`.

Subsequent store updates from the SSE handler are lost, and the agent panel on the right side remains blank.

Before PR #1280, the guard condition was `initState !== 'done' || installationState === 'waiting-backend'`. This ensured that even if `initState` was `'done'` during a restart, the application would still wait for the backend health check before rendering the main UI. After PR #1280, the guard condition was simplified to `initState !== 'done'`, which causes the application to skip waiting for backend readiness and render the main UI immediately, even though `chatStore` may still be `null`.

nitpicker55555:
This pr also fix flash problem mentioned in #1280 
In the old implementation, the race condition occurred because installationState === 'waiting-backend' depended on the asynchronous invocation of setWaitingBackend(). During the first render frame, the state was still 'idle', so the guard did not block execution.

By contrast, !isBackendReady evaluates to true from the very first frame (since its initial value is false), so this approach does not introduce a race condition.

PR #1350, opened 2026-02-23
enhance: browser_external_cdp PR696

Wendong-Fan:
two fixs:

Health-check now probes a snapshot of the pool and removes dead entries by id instead of array index, this avoids race issues when IPC add/remove operations happen while health-check is awaiting network probes

The UI now checks the result of addCdpBrowser before showing success, if adding to the pool fails, it shows the error instead of incorrectly showing a success state


nitpicker55555:
Thanks! LGTM

PR #1474, opened 2026-03-09
fix: show trigger action after markdown finishes typing

### LINKED ISSUE #1405: [Feature Request]  ![](https://files.slack.com/files-pri/T06R8AF1H46-F0AHY7AR3FY/screenshot_2026-03-02_at_14.55.44.png)
### Solution

_No response_

### Alternatives

Image

The current trigger action experience feels disconnected from the chat output. If the “Add trigger action” box appears before the markdown result has fully finished typing, it can interrupt the reading flow and make the response feel visually premature. This creates a less polished UX and may reduce clarity around when the user is expected to take action.

### Additional context

UX rendering logic

Update the chatbox rendering behavior so the “Add trigger action” box only appears after the markdown result has finished typing.

This ensures:
- The user can fully read the generated response first
- The follow-up action feels contextual, instead of interruptive
- The overall chat experience feels more natural and polished

Expected behavior:
- While the markdown response is still typing, the trigger action box should remain hidden
- Once the markdown typing/rendering is complete, the trigger action box should appear below the final response
- The trigger box should be treated as a post-response action, not part of the streaming content

kiannidev:
Fixes #1405.

## Summary

- Ensure the “Add trigger action” card (TaskCompletionCard) only appears **after** the assistant’s final END message has fully finished streaming/typing.
- Add a per-query `isCompletionReady` flag in `UserQueryGroup` that is set only when the END `AgentMessageCard`’s typewriter completes.
- Gate `TaskCompletionCard` rendering on both `task.status === 'finished'` and `isCompletionReady`, so the UI no longer shows trigger actions while the markdown answer is still typing.

## Implementation details

- Updated `src/components/ChatBox/UserQueryGroup.tsx`:
  - Track `isCompletionReady` with local state and reset when the active task or query group changes.
  - Pass `onTyping` to the END `AgentMessageCard` to set `isCompletionReady` when its `MarkDown` typewriter finishes.
  - Render `TaskCompletionCard` only when `task.status === 'finished' && isCompletionReady`.

## Testing

- Started a new project and submitted a task that produces a multi-line, streaming markdown answer.
- Verified that:
  - During streaming, the “Add trigger action” card does **not** appear.
  - As soon as the END message finishes typing, the “Add trigger action” card appears below the final answer.
- Replayed an existing project and confirmed the same behavior holds for replayed conversations.


4pmtong (approved):
LGTM. Thanks @kiannidev for contribution

PR #1536, opened 2026-04-02
refactor: establish Brain-centered architecture and frontend/backend separation foundations

4pmtong:
### Related Issue

Closes #

### Description

This PR delivers the current Brain + Web separation milestone.

It continues the ongoing refactor toward a Brain-centered architecture, where runtime capabilities are determined by the Brain environment rather than by client type. The goal of this phase is to preserve the existing Desktop experience while making `Web + Local Brain` a supported path built on shared, Brain-facing abstractions instead of Electron-specific assumptions.

Within that scope, this PR has three primary outcomes:

- it strengthens the Brain-side architecture around routing, session handling, Hands resolution, capability detection, and resource abstraction
- it advances frontend separation so shared web/desktop flows use Host and Brain-facing paths instead of depending directly on Electron-only globals
- it folds several high-impact regressions discovered during the refactor back into the new architecture, so the milestone improves behavior without adding more legacy branching

Concretely, this PR moves the codebase further toward the target architecture described in the refactor work:

- Brain capabilities are determined by deployment environment, not by client channel
- shared frontend code should not depend directly on Electron-only globals
- browser, file, and session flows should be routed through Brain / Hands abstractions
- Desktop remains backward compatible while Web gains first-class support

Notable user-facing improvements included in this PR:

- web login now returns users to the web app instead of redirecting into an Electron-only flow
- `Browser > CDP Browser Connection` now works in `dev:web` by using Brain APIs in web mode
- browser connection handling is normalized around endpoint-based state, keeping Local Brain as the default behavior today while preparing the path for future Hands-based browser resource expansion
- task execution now reuses an existing connected CDP endpoint instead of resetting browser state unnecessarily
- Agent Folder rendering now handles duplicate filenames correctly when multiple tasks generate files with the same name

This PR should be reviewed as a refactor milestone rather than as a collection of isolated bug fixes. The fixes included here are intentionally integrated into the Brain/Web separation work so the resulting behavior stays aligned with the target architecture.

Out of scope for this PR:

- making RemoteHands the default runtime path
- changing the default Local Brain deployment model
- completing future cluster / worker orchestration beyond the extension points needed for this refactor phase

### Testing Evidence (REQUIRED)

- [ ] I have included human-verified testing evidence in this PR.
- [ ] This PR includes frontend/UI changes, and I attached screenshot(s) or screen recording(s).
- [ ] No frontend/UI changes in this PR.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [x] Other


backend/app/hands/capabilities.py
```diff
@@ -0,0 +1,236 @@
...
+    Each field maps to a Hand Type: what the Brain can operate.
+    """
+
+    has_terminal: bool = True
+    """terminal hand: can execute shell"""
+
+    has_browser: bool = False
+    """browser hand: can control CDP browser"""
+
+    filesystem_scope: str = "full"
```
bytecii, line 108: Maybe make full etc as the strenum types?

backend/app/hands/capabilities.py
```diff
@@ -0,0 +1,236 @@
...
+    """mcp hand: all | allowlist"""
+
+    mcp_allowlist: list[str] = field(default_factory=list)
+    """used when mcp_mode=allowlist"""
+
+    workspace_root: str = "~/.eigent/workspace"
+    """workspace root path"""
+
+    deployment_type: str = "local"
+    """deployment type (for logging): local | cloud_vm | sandbox | docker"""
```
bytecii, line 123: what's the cloud vm? just curious

backend/app/hands/capabilities.py
```diff
@@ -0,0 +1,236 @@
...
+    deployment = env("EIGENT_DEPLOYMENT_TYPE") or ""
+    deployment = deployment.lower().strip()
+
+    if deployment in DEPLOYMENT_FULL:
+        # local/cloud VM -> full capabilities
+        in_docker = _is_running_in_docker()
+        if in_docker:
+            logger.info("Brain running in Docker, using limited capabilities")
+            deployment = "docker"
+            caps = BrainCapabilities(
```
bytecii, line 153: QQ should this be the HandCapabilities instead? for example browser is hand's capability?

backend/app/hands/sandbox_hands.py
```diff
@@ -0,0 +1,94 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2025-2026 @ Eigent.ai All Rights Reserved. =========
+
+from pathlib import Path
+
+from app.hands.interface import IHands
+
+# Terminal command allowlist for sandbox
+SANDBOX_TERMINAL_ALLOWLIST = frozenset(
```
bytecii: Seems that this is unused?

backend/app/hands/cluster_interface.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Any
+
+
+class IHandsCluster(ABC):
+    """Remote Hands worker cluster interface placeholder."""
+
+    @abstractmethod
+    async def acquire(
+        self,
+        resource_type: str,
```
bytecii, line 25: Make resource type to string enum?

backend/app/hands/remote_hands.py
```diff
@@ -0,0 +1,136 @@
...
+
+import asyncio
+from pathlib import Path
+from typing import Any
+
+from app.hands.cluster_interface import IHandsCluster
+from app.hands.interface import IHands
+
+
+class RemoteHands(IHands):
```
bytecii, line 23: Maybe we can put all Hands to a hands subfolder and import it to the hands/__init__.py

backend/app/hardware/interface.py
```diff
@@ -0,0 +1,55 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2025-2026 @ Eigent.ai All Rights Reserved. =========
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class IHardwareBridge(ABC):
+    """Hardware capability bridge. Only Desktop has implementation, others use NullBridge"""
```
bytecii, line 20: So should we implement this for the desktop?

backend/app/router_layer/hands_resolver.py
```diff
@@ -0,0 +1,206 @@
...
+    IHandsCluster,
+    RemoteHands,
+    RoutedHandsCluster,
+    SandboxHands,
+    load_hands_cluster_config,
+)
+from app.hands.capabilities import detect_capabilities
+from app.hands.environment_hands import EnvironmentHands
+
+logger = logging.getLogger("hands.resolver")
```
bytecii: Let's use the following
```suggestion
logger = logging.getLogger(__name__)
```

backend/app/controller/chat_controller.py
```diff
@@ -184,8 +261,15 @@ async def post(data: Chat, request: Request):
     if safe_env_path:
         load_dotenv(dotenv_path=safe_env_path)
 
+    # TODO(multi-tenant): os.environ is global – concurrent sessions overwrite
+    # each other's API keys, file paths, and browser ports.  Pass these values
+    # through Chat / request context instead of mutating the process environment.
     os.environ["file_save_path"] = data.file_save_path()
```
bytecii, line 275: to my understanding as the brain in the server, will it has some problems if we set env as previously just in desktop?

bytecii (commented):
In general the design is ok. But IMO we should refactor and put things to different PRs, otherwise it's hard to review.

backend/app/controller/chat_controller.py
4pmtong: Yes, It need redesign if we deploy the brain in the server, can use S3 instead of it.

backend/app/agent/tools.py
```diff
@@ -79,6 +93,24 @@ async def get_toolkits(tools: list[str], agent_name: str, api_task_id: str):
...
+                    )
+                    continue
+                if (
+                    item in BROWSER_DEPENDENT_TOOLKITS
+                    and not hands.can_use_browser()
+                ):
+                    logger.info(
+                        f"Skipping {item} for {agent_name}: no browser hand"
+                    )
+                    continue
```
a7m-1st, line 113: I think we can simplify this?

backend/app/hands/capabilities.py
```diff
@@ -0,0 +1,236 @@
...
+    """filesystem hand: full | workspace_only | none"""
+
+    mcp_mode: str = "all"
+    """mcp hand: all | allowlist"""
+
+    mcp_allowlist: list[str] = field(default_factory=list)
+    """used when mcp_mode=allowlist"""
+
+    workspace_root: str = "~/.eigent/workspace"
+    """workspace root path"""
```
a7m-1st: Can we convert this to PATH type instead? Resolving the path might cause issues down the line.

backend/app/controller/file_controller.py
```diff
@@ -0,0 +1,312 @@
...
+
+@router.get("/files")
+async def list_project_files(
+    project_id: str = Query(..., description="Project ID"),
+    email: str = Query(..., description="User email"),
+    task_id: str | None = Query(
+        None, description="Optional task ID to scope listing"
+    ),
+) -> list[dict]:
+    """
```
a7m-1st, line 200: Not sure if its me; my task was to write "hello world" file. The file is generated, but the route returned 
`/files/stream?path=task_1776131785382-755%5Chello_world.txt&project_id=1776121925475-837&email=ahmed%40gmail%20com`
- Thus if you try to click the file title in Eigent, it doesn't open the file. The file rendering is not there too.
- Lastly just a small comment, previously we were fetching "All project files", thus it appeared having flat hierarchy. But now they are nested ; just a small comment.

Image

backend/app/file_access/upload_file_access.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class UploadFileAccess(IFileAccess):
+    """Only handle uploaded files, path restricted to workspace/{session_id}/ (Web/Channel)"""
+
+    def __init__(self, workspace_root: str = "~/.eigent/workspace") -> None:
+        self.workspace_root = Path(workspace_root).expanduser()
+
+    def read_file(self, path: str) -> str:
+        resolved = self._ensure_in_workspace(path)
```
a7m-1st, line 27: Actually, a neat idea (if we don't mind the bandwidth) is to just send base64 strings as attachments. Previously the /chat sends attaches[] with local file paths. Now it is:
```
1. upload file separately                                                   
2. get file_id                                                              
3. send the message with file references
```
We can instead attach a base64 string, then decode it back then save it in whatever the backend is hosted (vm, docker or others) or upload it in the background. This way we don't need to have separate steps; for <10mb resources.
ref: https://github.com/a7m-1st/medigent/blob/main/backend/app/utils/file_utils.py

backend/app/file_access/local_file_access.py
```diff
@@ -0,0 +1,57 @@
...
+            p.write_text(content, encoding="utf-8")
+        else:
+            p.write_bytes(content)
+
+    def exists(self, path: str) -> bool:
+        return Path(path).exists()
+
+    def list_dir(self, path: str) -> list[str]:
+        return [p.name for p in Path(path).iterdir()]
+
```
a7m-1st, line 45: Rereading from disk (recursively) using python is expensive. Especially while SSE running, the python thread gets blocked thus waiting for some time before response. We mitigated this issue by migrating to websockets: ref: https://github.com/a7m-1st/medigent/blob/main/backend/app/controller/session_controller.py

Also Consider caching the get all files response unless there is a change.
ref: https://github.com/a7m-1st/medigent/blob/c32f3cf8cad1daa4fccb80085a9668286628f2f8/backend/app/controller/chat_controller.py#L453

a7m-1st: For the first issue; there should be a better solution.. websockets just supressed the problem it didn't fix it.

backend/app/hands/capabilities.py
a7m-1st: I think she meant a full virtual machine remotely; not just in docker.

4pmtong:
Thanks all for the detailed review. @bytecii @a7m-1st  I addressed the actionable, low-risk items in this PR:

1. Fixed file link generation for Windows-style separators (`\`) by normalizing relative paths to POSIX style before URL encoding, so `/files/stream` links open/render correctly.
2. Changed `workspace_root` in `BrainCapabilities` to `Path` (instead of `str`) to reduce downstream path-resolution risk.
3. Removed an unused constant in `sandbox_hands.py`.
4. Updated logger initialization in `hands_resolver.py` to `logging.getLogger(__name__)`.

For broader architecture suggestions (enum refactor, larger file-transfer redesign, filesystem caching/perf strategy, and env-scoping redesign for multi-tenant server mode), I suggest handling them in follow-up PRs to keep this refactor PR focused and reviewable.

PR #1622, opened 2026-05-09
Keep replay and history prompts tied to the right task

### LINKED ISSUE #1366: [BUG] task question only the first one loaded if reentered the project

0.0.86

### System information

0.0.86

### Problem description

When loading a project with multiple tasks from history, all tasks were incorrectly shown with the first task's question.

Image

### log

N/A

### Additional context

_No response_

### LINKED ISSUE #1367: [BUG] Replay shows wrong question

0.0.86

### System information

N/A

### Problem description

When clicking Replay on a specific task, it was using the earliest question found across all chat stores (always task 1's) instead of the active task's own question.

### log

N/A

### Additional context

_No response_

09Catho:
This fixes two closely related prompt mixups in the project replay/history flow.

When replaying a finished task, the app could grab the first prompt it found anywhere in the project instead of the prompt that belonged to the task you actually replayed. Reopening a multi-task project from history had the same shape of bug: every restored task could inherit one shared question instead of its own original prompt.

What changed:
- replay now reads the active task's own first user message
- history loading now carries a per-task question map so each restored task keeps its own prompt
- the history entry points were updated to pass those task-specific prompts through
- regression tests were added for both the replay path and multi-task history reload path
- the relevant test mocks were refreshed so those regression tests can actually run

Closes #1367
Closes #1366

Tested with:
- `npx vitest run test/integration/chatStore/replayComplete.test.tsx -t "active task question|per-task questions|correct first question on replayActiveTask"`
- `npm run type-check`

### DIFF
--- a/src/components/GroupedHistoryView/ProjectGroup.tsx
+++ b/src/components/GroupedHistoryView/ProjectGroup.tsx
@@ -134,6 +134,11 @@ export default function ProjectGroup({
         const taskIdsList = project.tasks
           ?.map((t) => t.task_id)
           .filter(Boolean) || [project.project_id];
+        const taskQuestions = Object.fromEntries(
+          (project.tasks || [])
+            .filter((task) => task.task_id && task.question)
+            .map((task) => [task.task_id, task.question])
+        );
 
         setIsLoadingProject(true);
         try {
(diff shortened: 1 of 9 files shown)


Douglasymlai:
Hi @09Catho , thanks for your contribution.

We’ve recently released a new version of Eigent with several feature updates, and the issue addressed in this PR has already been fixed by our internal team.

I’ll close this PR for now, but we really appreciate your effort and would be happy to see your contributions on other issues in the future.

PR #1636, opened 2026-05-14
fix: use urlencode for OAuth authorize URL parameters

hobostay:
## Summary
- Replace f-string URL construction with `urlencode()` in Slack, Notion, X (Twitter), and Google OAuth adapters (`server/app/core/oauth_adapter.py`)
- Parameters like `redirect_uri`, `scope`, and `state` could contain special characters (spaces, ampersands, Unicode) that would break the URL when interpolated directly
- LinkedIn adapter was already using `urlencode` correctly — this brings the other adapters in line
- X adapter's manual `self.scope.replace(' ', '%20')` is now handled properly by `urlencode`

## Test plan
- [ ] Verify OAuth login flows still work for Slack, Notion, X, Google, and LinkedIn
- [ ] Test with redirect URIs containing special characters
- [ ] Verify scope strings with spaces are properly encoded

🤖 Generated with [Claude Code](https://claude.com/claude-code)


4pmtong (approved):
Looks good to me. Thanks @hobostay for contribution

PR #1664, opened 2026-06-08
chore: Fix provider configuration state

### LINKED ISSUE #1665: BYOK provider stays unconfigured after successful validation
## Problem

A user can enter valid BYOK/custom provider settings, receive a successful validation result, but the provider still appears as not configured or old values are restored after refresh.

This was reported for both AWS Bedrock and AWS Bedrock Converse, but the issue can affect all custom/BYOK providers.

## Symptoms

- Validation can return success.
- The provider card may still show `Not Configured`.
- Updated fields such as endpoint URL, model type, or Bedrock Converse region may revert after saving.
- Newly saved providers may fail to become the default immediately.

## Root Cause

The frontend and server provider state fields are inconsistent: the frontend used `is_valid`, while the server stores `is_vaild`. The provider update path also failed to persist some fields needed by BYOK providers, and the frontend could read a stale `provider_id` immediately after creating a provider.

## Expected Behavior

After successful validation and save, the provider should remain configured, persist the entered values, and be selectable as the default model.

## Related PRs

- App/frontend fix: https://github.com/eigent-ai/eigent/pull/1664
- Server fix: https://github.com/eigent-ai/eigent_server/pull/112

fengju0213:
## Summary

Fixes the provider configuration flow so a successfully validated BYOK/custom provider is saved, refreshed, and marked configured correctly.

Fixes #1665

## Root Cause

The frontend sent `is_valid`, while the server model uses the existing `is_vaild` field. The server update allowlist also omitted `endpoint_url`, `model_type`, and `encrypted_config`, so updates could appear successful while leaving the stored provider unchanged. New providers could also fail to become the default immediately because the frontend read `provider_id` from stale React state.

## Changes

- Send and read the provider validity state using the server-compatible `is_vaild` field, while still accepting `is_valid` for compatibility.
- Allow provider updates to persist `endpoint_url`, `model_type`, and `encrypted_config`.
- Use the freshly loaded provider id when setting a newly saved provider as default.
- Add server coverage for the `is_valid` compatibility input.

## Impact

This affects all custom/BYOK providers, including AWS Bedrock and AWS Bedrock Converse. Direct AWS Bedrock users should still use the Bedrock Converse provider; the plain Bedrock provider remains for OpenAI-compatible Bedrock gateways.

## Validation

- `npm run type-check`
- `npx eslint src/pages/Agents/Models.tsx --no-warn-ignored --quiet`
- `uv run --extra dev pytest tests/test_provider_model.py`
- `uv run ruff check app/model/provider/provider.py app/domains/model_provider/service/provider_service.py tests/test_provider_model.py`

### DIFF
--- a/server/app/domains/model_provider/service/provider_service.py
+++ b/server/app/domains/model_provider/service/provider_service.py
@@ -14,9 +14,9 @@
 
 """ProviderService: model provider CRUD with prefer/invalidate. Follows CreditsService pattern."""
 
-from sqlalchemy import update
-from sqlmodel import select, col
 from loguru import logger
+from sqlalchemy import update
+from sqlmodel import col, select
 
 from app.core.database import session_make
 from app.model.provider.provider import Provider, VaildStatus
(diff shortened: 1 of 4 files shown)


Douglasymlai:
This issue has been fixed and moved to server side. Will close this PR for now.

PR #1674, opened 2026-06-09
fix: replayActiveTask uses active task question instead of earliest across all tasks (fixes #1367)

### LINKED ISSUE #1367: [BUG] Replay shows wrong question

carlosjarenom:
## Problem

When clicking Replay on any task, eigent always used the question from **task 1** (the earliest task across all chat stores), instead of the question from the currently active task.

This was reported by a maintainer in issue #1367.

## Root Cause

`replayActiveTask()` in `src/lib/replay.ts` was iterating over ALL chat stores and tasks to find the earliest user message by timestamp, rather than using the active task's own messages.

## Fix

Replaced the entire `earliestTimestamp` search logic with a direct lookup in the active task's messages.

## Testing

1. Create a project with multiple tasks
2. Complete task 1 and task 2 with different questions
3. Click Replay on task 2
4. Verify that the replay uses task 2's question, not task 1's

## Files Changed

- `src/lib/replay.ts` - Fixed `replayActiveTask()` to use active task's own question


Douglasymlai:
Hi @carlosjarenom, could you help resolve the conflicts?

Once they’re resolved, I can move this forward to the review process.

carlosjarenom:
Rebased on latest upstream/main and force-pushed. Conflicts resolved.

**Changes:** Replaced `replayActiveTask()` in `replay.ts` — was iterating all chat stores/tasks to find the earliest user message by timestamp, now directly uses the active task's own messages. Uses `replayProject()` instead of deprecated `chatStore.replay()`.

Douglasymlai:
@carlosjarenom Thanks for working on this. The Replay function has currently been removed from the product and will be brought back in the future, although some legacy Replay code remains in the codebase.

@4pmtong Could you review whether this corrected active-task lookup would be useful to retain now for the future Replay implementation? The helper is not currently connected to a live UI flow, and the branch presently fails TypeScript because ChatStore and ProjectStore are undefined, so those types would still need to be updated before merging.

4pmtong:
@carlosjarenom  Thanks for working on this. The idea of taking the replay question from the active task makes sense.

A few things to fix before merge:

- `npm run type-check` fails because `ChatStore` and `ProjectStore` are used in `src/lib/replay.ts` but not imported.
- `replayActiveTask` does not appear to be used by any production code path yet, so the user-facing replay flow may still be unchanged.
- Passing the current `projectId` into `projectStore.replayProject` can overwrite the existing project via `removeProject(projectId)`, which may also remove Space metadata. For active-task replay, we should use a separate replay project id or a non-destructive replay path.

Thanks again for the contribution.

PR #1676, opened 2026-06-09
fix: add loop="infinite" to img tags to ensure GIFs play continuously (fixes #1480)

### LINKED ISSUE #1480: [BUG] The generated GIF only plays once and then stops.

0.0.87

### System information

mac m2

### Problem description

The generated GIF only plays once and then stops.

Image

### log

[eigent-0.0.87-darwin-arm64-1773180089141.log](https://github.com/user-attachments/files/25884651/eigent-0.0.87-darwin-arm64-1773180089141.log)

### Additional context

_No response_

carlosjarenom:
## Problem

GIFs generated by eigent agents would only play once and then stop on the last frame, instead of looping continuously.

Reported in issue #1480.

## Fix

Added `loop="infinite"` attribute to all `` tags in the MarkDown component (`src/components/ChatBox/MessageItem/MarkDown.tsx`), covering:

- Images rendered from relative paths (converted to data URLs)
- Images rendered from absolute URLs
- Image preview dialog

While HTML `` tags should loop GIFs by default, this explicit attribute ensures consistent behavior across browsers and Electron.

## Files Changed

- `src/components/ChatBox/MessageItem/MarkDown.tsx` - Added `loop="infinite"` to img tags


carlosjarenom:
Rebased on latest upstream/main and force-pushed.

**Changes:** Added `loop="infinite"` to all `` tags in `MarkDown.tsx` (relative paths, absolute URLs, preview dialog) — ensures GIFs loop continuously across browsers and Electron.

Douglasymlai:
@carlosjarenom Thanks for the contribution. This implementation does not fix #1480. The loop attribute is not valid for an HTML  element, and Chromium/Electron will ignore it. GIF repetition is defined within the GIF data itself, so adding loop="infinite" cannot make a play-once GIF repeat.

The branch also currently fails TypeScript and ESLint checks because of the unsupported attribute. Please fix the generated GIF data to include infinite looping, or use another supported mechanism to restart the animation. A regression test using a known one-shot GIF would also be helpful.

@4pmtong Could you also take a look at the expected implementation for this fix?

4pmtong:
Thanks @Douglasymlai, I agree with this assessment.

From our current docs/design direction, I would avoid fixing this in `MarkDown.tsx`. The markdown renderer should stay a display layer; generated artifact behavior should be fixed either at the GIF generation step by writing infinite-loop metadata, or through a supported viewer-level replay/normalization path. That will also keep the fix compatible with the future WorkspaceResolver / ArtifactRef / remote-control flow.

carlosjarenom:
Superseded by #1780 — the fix for #1480 is implemented at the generation level (backend) rather than display level. Closing this PR.

PR #1677, opened 2026-06-09
fix: replace hardcoded credentials in docker-compose and harden server/.gitignore (#1527)

carlosjarenom:
## Summary

Fix remaining security issues from #1527 (part 2 was already addressed in a prior commit).

### 1. Replace hardcoded database passwords in compose files
- `server/docker-compose.yml`: Replace all instances of `123456` with `${POSTGRES_PASSWORD:-123456}`
  - `POSTGRES_PASSWORD` environment variable for the Postgres service
  - `database_url` references in `api`, `celery_worker`, and `celery_beat` services (both build args and runtime env)
- `server/docker-compose.dev.yml`: Same replacement for the Postgres service

### 2. Harden `server/.gitignore`
- Add `.\env` to prevent accidental credential commits from the server directory
- Add `alembic.ini` which may contain database connection strings

### 3. Update `server/.env.example`
- Add `POSTGRES_PASSWORD=CHANGE_ME` placeholder
- Change `secret_key=postgres` to `secret_key=CHANGE_ME`

## Backwards Compatibility

All substitutions use `${POSTGRES_PASSWORD:-123456}` syntax, so existing deployments that don't set `POSTGRES_PASSWORD` will continue to work with the default value. New deployments can override by setting the variable in their `.\env` file.

## Related

- Closes #1527 (partial — part 2 on chat_share.py was already fixed)

### DIFF
--- a/backend/app/agent/toolkit/terminal_toolkit.py
+++ b/backend/app/agent/toolkit/terminal_toolkit.py
@@ -388,7 +388,9 @@ class TerminalToolkit(BaseTerminalToolkit, AbstractToolkit):
         # Post-process GIF files to ensure infinite loop
         if self.working_directory and block:
             try:
-                modified = post_process_gifs_in_directory(self.working_directory)
+                modified = post_process_gifs_in_directory(
+                    self.working_directory
+                )
                 if modified:
                     logger.debug(
                         f"Post-processed {modified} GIF file(s) for infinite loop"
--- a/backend/app/agent/toolkit/terminal_toolkit.py
+++ b/backend/app/agent/toolkit/terminal_toolkit.py
@@ -35,6 +35,7 @@ from app.service.task import (
     get_task_lock,
     process_task,
 )
+from app.utils.file_utils import post_process_gifs_in_directory
 from app.utils.listen.toolkit_listen import auto_listen_toolkit
 
 logger = logging.getLogger("terminal_toolkit")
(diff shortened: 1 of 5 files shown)


carlosjarenom:
Rebased on latest upstream/main and force-pushed.

**Changes:** Replaced hardcoded `123456` passwords in `docker-compose.yml` and `docker-compose.dev.yml` with `${POSTGRES_PASSWORD:***}` env var syntax. Added `.env` and `alembic.ini` to `server/.gitignore`. Updated `.env.example` with `POSTGRES_PASSWORD=***` and `secret_key=CHANGE_ME`.

4pmtong:
Hi @carlosjarenom Thanks for the security cleanup here. One small thing to double-check: `database_url` is now passed through `build.args`, and `server/Dockerfile` currently turns `ARG database_url` into `ENV database_url=$database_url`. If someone sets a real `POSTGRES_PASSWORD`, that credential may end up baked into the Docker image metadata after build.

Could we keep `database_url` runtime-only in compose and avoid passing the deployment password as a build arg? If the build still needs a value, a non-secret placeholder would be safer.

carlosjarenom:
Thanks for flagging the `database_url` build-arg risk — you were right, that would have left the credential in the image metadata. Addressed in `7d2aa72`: we now pass a non-sensitive placeholder during build (`args: [DATABASE_URL=postgresql://postgres:placeholder@postgres:5432/eigent]`), and the real `DATABASE_URL` is injected at runtime via the `.env`-driven `environment` block. The `.env` file is gitignored via `server/.gitignore`, per the issue #1527 scope.

Pre-commit CI also flagged a few cosmetic issues (trailing-whitespace / ruff / ruff-format). Those are also fixed in `a547e8c` on top of the security changes. Happy to iterate if you'd prefer a different approach.

PR #1678, opened 2026-06-09
fix: use per-task question when loading from history (#1366)

### LINKED ISSUE #1366: [BUG] task question only the first one loaded if reentered the project

carlosjarenom:
## Summary

When loading a project from history, every task was replayed with the same question
(`project.last_prompt`), regardless of what question each individual task had been
created with. This caused tasks to show incorrect questions when replayed.

## Changes

- `src/store/projectStore.ts`: Added `taskQuestions?` parameter to `loadProjectFromHistory`;
  each task now uses `taskQuestions[index] || question` instead of always using `question`
- `src/lib/replay.ts`: Pass `taskQuestions?` through to `projectStore.loadProjectFromHistory`
- `src/components/HistorySidebar/index.tsx`: Extract `taskQuestionsList` from `project.tasks`
  (each task has its own `question` field) and pass it to `loadProjectFromHistory`

## Why

Each `HistoryTask` already stores its own `question` field. This fix ensures that field
is actually used during history replay, falling back to the project-level question only
when a task's question is missing.

Closes #1366

### DIFF
--- a/src/components/HistorySidebar/index.tsx
+++ b/src/components/HistorySidebar/index.tsx
@@ -146,6 +146,9 @@ export default function HistorySidebar() {
     const taskIdsList = project?.tasks.map(
       (task: HistoryTask) => task.task_id
     ) || [projectId];
+    const taskQuestionsList = project?.tasks.map(
+      (task: HistoryTask) => task.question
+    );
 
     // If no tasks to replay, create an empty project
     if (!taskIdsList || taskIdsList.length === 0) {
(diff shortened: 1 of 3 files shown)


Douglasymlai:
Hi @carlosjarenom , thanks for your contribution.

We’ve released a new version, and the issue covered in this PR has now been resolved. I’ll close this PR for now, but we appreciate your time and contribution.

PR #1703, opened 2026-06-22
fix: keep .git in build context so server Docker build succeeds

### LINKED ISSUE #1702: [BUG] docker compose build fails: `.git` not found in build context (regression from 1.0.0)
### Summary

Following `server/README_EN.md` step `docker-compose up --build -d` (from a clean state, no cached layers) fails on the API image build:

```
> [stage-0 14/14] RUN --mount=type=bind,source=.git,target=/tmp/.git \
      echo "EIGENT_SERVER_GIT_COMMIT=$(git --git-dir=/tmp/.git log -1 --format=%H -- server/ 2>/dev/null || echo unknown)" > /app/.image_env:
------
failed to solve: failed to compute cache key: failed to calculate checksum of ref ...: "/.git": not found
```

### Root cause

A regression introduced in the 1.0.0 release.

- `server/Dockerfile:66` (added in PR #1517, March 2026) bind-mounts `.git` from the build context to stamp the server commit into the image:
  ```dockerfile
  RUN --mount=type=bind,source=.git,target=/tmp/.git \
      echo "EIGENT_SERVER_GIT_COMMIT=$(git --git-dir=/tmp/.git log -1 ...)" > /app/.image_env
  ```
- `.dockerignore` was added new in commit 97d75544 (release: Eigent 1.0.0, PR #1695) and excludes `.git`:
  ```
  # Source control
  .git
  .gitignore
  .gitattributes
  ```
- BuildKit honors `.dockerignore` for `--mount=type=bind` sources, so the bind mount cannot find `.git` in the context and the build fails.

This was likely not caught pre-merge because cached image layers from before `.dockerignore` was added masked the failure on maintainer machines. A fresh clone + first build hits it immediately.

### Reproduction

```bash
git clone https://github.com/eigent-ai/eigent.git
cd eigent/server
cp .env.example .env
docker compose build --no-cache api   # or `docker compose up --build -d`
```

### Environment

- macOS (Darwin 25.5.0), Docker Desktop with BuildKit enabled (default)
- Tag: 1.0.0 (commit 97d75544)

### Fix

Stop ignoring `.git` in `.dockerignore`. The bind mount means `.git` is never baked into a layer, only made available during the build step. PR to follow.

nilbot:
## Summary

- The 1.0.0 release (#1695) introduced a new `.dockerignore` at the repo root that excludes `.git`.
- `server/Dockerfile` uses `--mount=type=bind,source=.git,target=/tmp/.git` (added in #1517) to stamp the latest `server/` commit into the image without baking `.git` into a layer.
- BuildKit honors `.dockerignore` for bind-mount sources, so fresh `docker compose build` runs fail with `failed to calculate checksum of ref ...: "/.git": not found`. Cached layers on maintainer machines masked this; new clones hit it immediately.

This PR removes the `.git` entry from `.dockerignore` and adds a comment explaining why it must stay in the build context. `.git` is never copied into any image layer (the bind mount is only available during the `RUN` step), so the only cost is a slightly larger build context — which BuildKit handles efficiently.

Fixes #1702

## Test plan

- [x] `docker compose build --no-cache api` — fails on `main`, succeeds with this patch
- [x] `docker compose up --build -d` — all containers healthy (`eigent_api`, `eigent_celery_worker`, `eigent_celery_beat`, `eigent_postgres`, `eigent-redis`)
- [x] `curl http://localhost:3001/health` → 200
- [x] `EIGENT_SERVER_GIT_COMMIT` is correctly written to `/app/.image_env` during build


MKrale (commented):
I've tested this change, and it indeed solves Bug #1702.

4pmtong (approved):
Looks good to me, Thanks @nilbot for contribution!

PR #1732, opened 2026-07-04
keep caret from jumping to start when typing fast

iyernaveenr:
### Problem
Typing quickly in the chat input makes the cursor jump to the start, so characters land at the beginning.

### Root cause
`RichChatInput` re-tokenizes and reassigns the contentEditable's `innerHTML` on every keystroke, then restores the caret inside a `requestAnimationFrame` callback. Reassigning `innerHTML` collapses the selection to offset 0; deferring the restore to the next frame leaves a window where fast keystrokes are inserted at position 0.

### Fix
Restore the caret synchronously immediately after updating the content, and defer only the layout-dependent `scrollCaretIntoView`. Rich-text/link highlighting and caret preservation are unchanged.

### Testing
Type rapidly in the chat input — the caret stays at the insertion point instead of jumping to the start.


Douglasymlai:

Douglasymlai (approved):

PR #1733, opened 2026-07-04
scroll down to latest msg when switching chats

iyernaveenr:
### Problem
Switching to a different project in the sidebar leaves the chat scrolled to the top instead of showing the latest message at the bottom.

### Root cause
In `ProjectChatContainer`, the auto-scroll effect only fires when a new user message is appended (the message count increases and the last message is from the user). The other scroll effects only handle explicit sidebar navigation. Nothing scrolls the viewport when the active project changes, so a switch renders the newly selected project's messages but keeps the previous scroll position (the top).

### Fix
Add an effect keyed on the active project id that scrolls the chat container to the bottom after a short defer (so the newly rendered messages have laid out). It uses instant scrolling so switching does not visibly scroll through the whole history, and it does not affect in-conversation scrolling or the existing "scroll to bottom on new user message" behavior.

### Testing
Switch between projects in the left sidebar - each now opens at the latest message (bottom) instead of the top.


Douglasymlai:

Douglasymlai (approved):

PR #1735, opened 2026-07-04
Fix invisible chat code block text when OS color scheme and app theme disagree

### LINKED ISSUE #1730: [BUG] Light mode markdown text/background colors invisible when OS is in dark mode

1.0.1

### System information

N/A

### Problem description

Image
### Description
When the app's in-app theme is manually set to Light mode, markdown content (code blocks, inline code, etc.) renders with invisible/incorrect text and background colors. Switching the OS-level system appearance to Light mode fixes it, confirming the app's own light-mode setting isn't what's driving the markdown colors.

### Root cause
`src/style/markdown-styles.css` imports `github-markdown-css/github-markdown.css` (see `node_modules/github-markdown-css/github-markdown.css` lines 14-125). That stylesheet only defines the CSS custom properties markdown-body relies on (`--fgColor-default`, `--bgColor-default`, `--bgColor-neutral-muted`, `--color-prettylights-syntax-*`, etc.) **inside** `@media (prefers-color-scheme: dark)` / `@media (prefers-color-scheme: light)` blocks:

​```css
@media (prefers-color-scheme: light) {
  .markdown-body, [data-theme="light"] {
    --fgColor-default: #1f2328;
    --bgColor-default: #ffffff;
    ...
  }
}
​```

These media queries key off the **OS-level** color scheme preference, not this app's own `data-theme` attribute. `src/style/markdown-styles.css` only re-declares `color-scheme: light|dark` for `[data-theme='light'|'dark'] .markdown-body` (lines 4-9) but never re-declares the actual color variables outside the media query.

So: if the OS is in dark mode but the user switches the in-app theme to Light, `@media (prefers-color-scheme: light)` doesn't match, `[data-theme="light"]` never gets its variables set, and every rule using `var(--fgColor-default)` / `var(--bgColor-muted)` etc. falls back to an unset/invalid value — making code blocks and other markdown elements render with mismatched or invisible colors. This only "looks correct" when the OS preference happens to agree with the in-app theme.

### Suggested fix
In `src/style/markdown-styles.css`, re-declare the full set of `--fgColor-*` / `--bgColor-*` / `--color-prettylights-syntax-*` / `--borderColor-*` variables directly under `[data-theme='light'] .markdown-body` and `[data-theme='dark'] .markdown-body` selectors (outside of any `@media` query), so the in-app theme always wins regardless of the OS setting.

### Affected files
- `src/style/markdown-styles.css`
- (upstream dependency for reference only, not to be edited: `node_modules/github-markdown-css/github-markdown.css`)

### log

N/A

### Additional context

_No response_

iyernaveenr:
Closes #1730

### Problem
In the chat, fenced code block text can render invisible (blank until selected). Originally reported with the OS in dark mode and the app in light theme, but the same failure exists in the mirrored combination (OS light, app dark).

### Root cause
`github-markdown-css` colors code text via `--fgColor-default`, keyed to the **OS color scheme**: its `@media (prefers-color-scheme: ...)` blocks apply the palette to bare `.markdown-body`, and `.markdown-body pre { color: var(--fgColor-default) }` picks it up (near-white when OS-dark, near-black when OS-light). The app's custom overrides in `src/style/markdown-styles.css` force code-block **backgrounds** per **app theme** (`data-theme`): light `#f6f8fa`, dark `#161b22`. Whenever the OS scheme and the app theme disagree, the OS-keyed text color lands on an app-keyed background of nearly the same color and becomes unreadable. Body prose is unaffected because a separate `.markdown-body { color: inherit }` override handles it, but that does not cascade into `pre`.

### Fix
Pin the code text color alongside each forced background, so text and background are always keyed to the same (app) theme:
- `.markdown-body pre` uses `color: #1f2328` next to its forced light background.
- `[data-theme='dark'] .markdown-body pre` uses `color: #e6edf3` next to its forced dark background.
- `.markdown-body pre code` uses `color: inherit` so block contents follow the block.

Inline `.markdown-body code` needs no pin: the stylesheet never sets a text color on it directly, so it correctly inherits the app-theme text color in both themes.

### Testing
Verified all four OS-scheme x app-theme combinations on a chat response containing prose, inline code, and a fenced code block:
- OS dark + app light (the reported bug): code text now readable.
- OS light + app dark: code text now readable.
- OS dark + app dark, OS light + app light: unchanged, still readable.


Douglasymlai:

Douglasymlai (approved):

4pmtong:
Looks good to me. Thanks @iyernaveenr for contribution!

PR #1736, opened 2026-07-04
Keep each chat session on its own model

iyernaveenr:
### Problem
Switching between chat sessions always shows (and sends with) the current global model, even when a session was started with a different model. There is no per-session model: the composer only ever reflects the global setting, and a follow-up in an old conversation silently runs on whatever the global model happens to be.

### Root cause
The model is resolved from the global `authStore` on every send (`startTask`), and the in-session composer's model indicator is derived from the same global state, so nothing stores or restores a per-session model. Sessions are durable at the Project level (each send creates a fresh task), so the model has to live on the Project.

### Fix
Persist the model a Project runs with in the Project's metadata (`metadata.modelSelection`) and reuse it for every later run:
- `projectStore`: new `setProjectModel` / `getProjectModel`. The selection is kept in the runtime store, mirrored into the persisted space meta, and also PATCHed into the server-side project metadata (shallow-merged), so it survives app restarts and space re-syncs from the server.
- `startTask`: when the Project has a stored selection, resolve the model from it -- custom/local providers are matched by their provider id (falling back to the default provider with a warning toast if the pinned one was removed), cloud models by the stored cloud model id, codex by the stored type. The resolved selection is re-captured after every run. The global cloud selection is no longer mutated when a per-project selection is in effect.
- Composer: the in-session model dropdown now shows the session's own model and is interactive -- changing it updates only that Project's selection (the global default and the server-wide preferred provider are untouched). The home/new-chat composer keeps its existing global-default behavior.

### Notes
- A selection change applies from the next run; an in-flight run keeps the model it started with.
- Unit tests cover the new store actions, including the restart path where the runtime store is empty and the selection is recovered from the persisted space meta.

### Testing
- `tsc` passes; the full test suite shows no regressions vs `main` (identical failure set); new store tests pass.
- Start a session on model A, switch the global model to B (home composer or another project), then send a follow-up in the first session: it runs on A and the composer shows A. Restart the app and follow up again: still A. Change the session's model from the in-session dropdown: the next run uses the new selection; other sessions and the global default are unaffected.


Douglasymlai:
Testing passed and the implementation works as described: each Project/session keeps its own model, and changing the model from the BottomBox after a run finishes applies to the next run without changing other Projects or the global default. Thanks for your contribution!

@4pmtong, could you please provide an additional review? I would also like a product/backend discussion on the intended behavior here: should an existing Project allow its saved model to be updated between runs so that each new run can switch to a different model? Please also review the backend metadata persistence and update flow before merging.

4pmtong (approved):
Look good to me, Thanks @iyernaveenr for contribution

PR #1740, opened 2026-07-06
Attach clipboard images pasted into the chat box

iyernaveenr:
### Problem
Pasting an image from the clipboard (e.g. a screenshot) into the chat input does nothing. Only plain text pastes work; attaching an image requires the file picker or drag-and-drop, which is a detour for the most common flow of sharing a screenshot with the agent.

### Root cause
The rich input's paste handler calls `preventDefault()` and then reads only `clipboardData.getData('text/plain')`; clipboard file items are never inspected, so pasted images are silently discarded. The attachment pipeline is path-based (`FileAttachment.filePath`), and unlike drag-and-dropped files, pasted `File` objects carry no filesystem path, so they could not simply be forwarded to the existing drop handler.

### Fix
Route pasted file items into the existing attachment flow:
- `RichChatInput` gains an `onPasteFiles` callback; its paste handler now extracts clipboard file items (`clipboardData.items` -> `getAsFile()`) and forwards them, while still inserting any plain text on the clipboard (some sources provide both).
- Desktop: a new `save-pasted-file` IPC handler persists the pasted bytes to a temporary directory and returns the path, since pasted files have no path of their own. A new `processPastedFiles` helper (mirroring `processDroppedFiles`) uses it and merges the result into the existing attachments, deduplicated by path.
- Web: pasted files are uploaded exactly like dropped files.
- The input box applies the same gating as drag-and-drop (attachments disabled in privacy-off / cloud-model-in-dev states) and shows the same success/error toasts.

All pasted file kinds are forwarded, so copying a file from a file manager and pasting it also attaches it.

### Testing
Verified on Linux desktop: copying a screenshot to the clipboard and pressing Ctrl+V in the chat box attaches it as a file chip (with toast), and the agent can read the attached image in the run; plain-text paste and drag-and-drop behave as before. `tsc` passes; lint passes for the changed files.


Douglasymlai:

electron/main/index.ts
```diff
@@ -1548,6 +1548,33 @@ function registerIpcHandlers() {
...
+        const filePath = path.join(pastedDir, `${stamp}-${safeName}`);
+        await fsp.writeFile(filePath, Buffer.from(new Uint8Array(data)));
+        return { success: true, filePath, fileName: safeName };
+      } catch (error: any) {
+        log.error('Failed to save pasted file:', error);
+        return { success: false, error: error.message };
+      }
+    }
+  );
+
```
4pmtong, line 1602: Could we make the saved pasted-file path unique per file? Clipboard images often share names like image.png, and the current seconds-level timestamp + filename can collide when multiple images are pasted in the same second.

4pmtong (changes requested):

electron/main/index.ts
iyernaveenr: Thanks for catching this -- you're right that a seconds-level timestamp plus a commonly shared name like `image.png` can collide when multiple images are pasted within the same second.

> Could we make the saved pasted-file path unique per file? ... the current seconds-level timestamp + filename can collide when multiple images are pasted in the same second.

Fixed -- the saved path now includes a random token (`crypto.randomUUID()`), so each pasted file gets a unique path even for same-second, same-name pastes. The name shown in the UI is unchanged (it still uses the original file name).

I've also brought the branch up to date with `main` (resolving a conflict from the recent bottom-bar redesign). Ready for another look when you have a chance.

4pmtong:
LGTM, Thanks @iyernaveenr !

4pmtong (approved):

PR #1745, opened 2026-07-06
Resume the last visited session at launch

iyernaveenr:
### Problem
Launching the app always lands on the workspace home ("Cowork with...") even when the user was in the middle of a conversation. Getting back to it requires finding and clicking the Project in the sidebar on every launch -- unlike editors that reopen where the user left off.

### Root cause
The pieces for a resume half-exist: the space store persists `activeSpaceId` and `lastVisitedProjectBySpace`, but the runtime project store is not persisted (`activeProjectId` is null at boot) and the workspace tab resets to its default. The existing fallback effect that activates the last visited Project only runs when the UI is already on a project-scoped tab, which is never the case at boot -- so nothing ever triggers the restore.

### Fix
On boot, when the persisted state records a last visited Project for the active Space, open it through the exact same path as clicking it in the sidebar (`selectProject`: history load, tab choice, achieved handling). Guards keep it polite:
- runs once per renderer boot (module-scoped flag, so sidebar remounts never re-trigger it);
- only from the pristine boot state (default tab and no active Project), so deliberately navigating home later is never hijacked;
- no-ops when there is no last visited entry, the Project is not shown in the nav list (e.g. archived), or on a fresh install.

All inputs to the decision are persisted locally, so the resume starts immediately at mount; the conversation content then loads the same way a sidebar click would load it.

### Testing
Verified on Linux desktop across full quit/start cycles: the app reopens directly in the last visited conversation, including after closing via the window button; clicking Home afterwards stays on home (no hijack); a Space with no last visited Project boots to home as before. `tsc` passes and lint is clean for the changed file.


Douglasymlai:

Douglasymlai (approved):

4pmtong:
Thanks @iyernaveenr LGTM!

PR #1746, opened 2026-07-06
Make the agent step timeout configurable and record spend of failed runs

### LINKED ISSUE #1692: [Feature Request] Improve task timeout behavior for long-running agent work
### Motivation

Eigent currently has fixed timeout behavior for long-running tasks. Some tasks can legitimately take longer than the current limit, especially complex tasks involving multiple agents or sub-agents.

A fixed timeout can stop a task too early even when the model is still making progress. This is especially problematic when a main agent is waiting for sub-agents to finish before continuing.

### Solution

Improve task timeout behavior so long-running tasks can be handled more flexibly.

The requirement is to avoid a hard-coded timeout that prematurely stops valid long-running work. The model/agent should have a better way to determine or request an appropriate timeout window based on the task context.

This should cover both product experience and backend behavior:
- Users should understand when a task is taking longer than expected.
- Backend task lifecycle should avoid stopping valid work too early.
- Main agent and sub-agent timeout behavior should be better aligned.

### Alternatives

Keep the current fixed timeout. This is predictable, but it can incorrectly terminate valid long-running tasks.

### Additional context

_No response_

### LINKED ISSUE #1718: [Feature Request] Improve task timeout design for long-running work
### Motivation

Some Eigent tasks can run longer than the current fixed timeout, especially complex tasks involving long execution, multiple agents, or sub-agents.

A fixed timeout can stop valid work too early. The main agent may also need to wait for sub-agents to complete before continuing, so timeout behavior should account for agent coordination instead of treating every layer the same way.

### Solution

Improve task timeout behavior and align it with competitor expectations.

The goal is to support long-running tasks more gracefully, avoid premature termination, and provide a better user experience when a task is taking longer than expected.

This issue tracks the requirement. The exact timeout policy and implementation details can be refined during development.

### Alternatives

Keep the current fixed timeout behavior. This is predictable, but it can incorrectly stop valid long-running tasks.

### Additional context

_No response_

iyernaveenr:
Closes #1718
Closes #1692

### Problem
Any agent turn longer than thirty minutes is aborted with `Async step timed out after 1800s`, regardless of whether it was making steady progress (observed with a long single-agent writing task that was still producing output when the ceiling hit). The limit is hardcoded, so deployments running long-form tasks cannot adjust it. In addition, a run that fails this way leaves its history row with zero tokens forever, hiding the real spend of the aborted work.

### Root cause
- `ListenChatAgent` hardcodes `step_timeout: float | None = 1800`; the underlying agent wraps each non-streaming step in `asyncio.wait_for` with that value, so the whole turn is cancelled at the ceiling. The screenshot toolkit repeats the same hardcoded fallback.
- On the frontend, the error step handler marks the task failed but never updates the run's history row, so `tokens` stays at its initial zero even though per-exchange token counts had been accumulated client-side.

### Fix
- Backend: read the default step timeout from `AGENT_STEP_TIMEOUT_SECONDS` (seconds; same 1800 default when unset; a non-positive value disables the per-step ceiling; invalid values log a warning and keep the default). The screenshot toolkit fallback uses the same source instead of a second hardcoded constant. Explicit `step_timeout` arguments (e.g. the short validation timeouts) behave as before.
- Frontend: when a run ends in an error, record the tokens accumulated so far into the run's history row (best-effort), so failed and timed-out runs no longer report zero spend.

### Testing
- Helper verified at runtime: unset -> 1800, `3600` -> 3600, `0` -> disabled, garbage -> warning + 1800. Both changed backend modules compile and import cleanly (no circular import).
- `tsc` passes; lint and formatting clean for the changed frontend file; store test suites pass.
- Reproduced the original failure with a long writing task aborted at exactly 1800s; with the token recording in place, an aborted run's history row carries the tokens consumed up to the failure.


4pmtong (approved):
Thanks @iyernaveenr LGTM!

PR #1747, opened 2026-07-06
Warn before closing the window while any run is still active

iyernaveenr:
### Problem
Closing the window while agent runs are streaming kills them silently: the event streams are severed, the backend treats it as a client disconnect and aborts the in-flight work, cleaning up the task locks. The affected conversations are left permanently incomplete (dangling tool rows, progress stuck, no final answer, zero recorded tokens) with no warning at close time.

### Root cause
A close warning exists, but it only checks the status of the active Project's current task. Runs streaming in other Projects -- exactly the situation when working with several Projects in parallel -- are invisible to it, so the window closes without the "a task is currently running" dialog.

### Fix
Track whether any run in any Project still holds a live event stream (the set of active stream controllers is the ground truth for what a window close would sever), and show the existing close-notice dialog when either the active task is busy or any such run exists. Cancel keeps the app and the runs alive; confirming still quits, making an interrupted run an informed choice instead of a silent side effect.

### Testing
Verified on Linux desktop: with a long-running task streaming in one Project and a different Project focused, closing the window now shows the running-task dialog (previously it closed silently and the backend logged a stream cancellation and aborted the run); Cancel left the runs untouched and they completed normally afterwards. With no runs active, the window closes immediately as before. `tsc` passes; lint and formatting are clean for the changed files.


Douglasymlai:
@iyernaveenr Thanks for the update. I found two issues during manual testing:

1. After the task finishes, closing the window still shows the "task is currently running" notice and prevents the window from closing normally. Please ensure the active-run state is cleared when a task finishes.
2. The close-window notice card should follow the same design as the delete/end-project confirmation dialog. Please align the title consistently with those dialogs and use the destructive/error button style for the "Yes" action, since confirming will terminate the running task.

Please update the PR to address these points.

iyernaveenr:
Thanks for the thorough manual testing and the clear write-up -- both points were spot on.

> closing the window still shows the "task is currently running" notice ... ensure the active-run state is cleared when a task finishes

Fixed -- a finished run's stream controller is now cleared on completion, so it no longer counts as active for the close guard.

> the close-window notice card should follow the same design as the delete/end-project confirmation dialog ... destructive/error button style

Done -- the notice now uses the shared confirm dialog (destructive confirm), so its title and button style match the end/delete-project dialogs.

Douglasymlai:
Looks good to me. Thanks for your contribution. I retested the updated close-window notice: the shared destructive confirmation style is now consistent, Cancel keeps the active run alive, and the notice no longer appears after the task finishes.

@4pmtong Please provide the second review when you have a chance.

4pmtong (approved):

PR #1748, opened 2026-07-07
Keep project switches from killing live runs

iyernaveenr:
### Problem
Switching to another Project and back while a run is streaming can kill the run: its stream is aborted client-side, the backend treats the disconnect as a cancel and aborts the in-flight work, and the conversation is left stuck as ongoing forever (spinner, no output, zero recorded tokens). To the user it looks like the run "never started".

### Root cause
When a Project is hydrated from a cached snapshot and the server has newer activity, the Project is flagged so its runtime state is evicted on the next transition away (forcing a fresh history load next time). That eviction also drops the chat stores of a still-streaming run. Returning to the Project then rebuilds it from history -- which includes the still-ongoing task -- and the history replay for that task id finds the live stream controller and aborts it ("already has an active SSE connection, aborting old one").

### Fix
- Skip the stale-cache eviction while any of the outgoing Project's runs still holds a live stream. The stale flag stays set, so the eviction (and fresh reload) simply happens on a later transition when nothing is running.
- Make history replay refuse to tear down a live stream in any case: if a replay targets a task id that still has an active stream, skip the replay -- the ongoing run is the fresher state.

### Testing
Reproduced deterministically before the fix (Linux desktop): open a Project whose cache is older than its server state, start a long-running task, switch to another Project and back -- the console logs the abort warning, the backend logs a stream cancellation, and the run dies with its session permanently stuck as ongoing. After the fix, the same sequence leaves the run untouched: no abort, no history reload, and the task completes normally with its output and token count recorded. `tsc`, lint, and the store test suites pass.


Douglasymlai:

4pmtong:
Hi @iyernaveenr Could we add regression coverage for this lifecycle? The fix depends on staleProjectIds + project transition + a live SSE controller all interacting correctly. A store-level test that keeps the outgoing project runtime while hasActiveSSEConnection(taskIds) is true, preserves the stale flag, and evicts on a later transition after the controller is gone would make this much safer to keep during future project/cache refactors.

4pmtong:
Thanks! I added the regression coverage. It now covers both sides of the lifecycle: keeping the stale outgoing Project runtime while live SSE is active, then evicting it on a later transition once the SSE is gone.

4pmtong (approved):

Issue #1578, opened 2026-04-24
[BUG] DeepSeek BYOK validation fails with "Unknown model platform: deepseek" (related to #1556)


davidemarchisio:
### What version of eigent are you using?

v0.0.90

### System information

Reproducible on Linux (and likely all platforms). Occurs when configuring DeepSeek as a BYOK custom model in the Models settings page.

### Problem description

Configuring DeepSeek as a BYOK provider fails at validation. The UI shows a validation error, and the backend logs:

```
Invalid model platform: deepseek
```

**Expected behavior:** DeepSeek should validate successfully (it's listed as a supported provider in the official docs at `docs/core/models/byok.md` with default API host `https://api.deepseek.com`).

**Actual behavior:** Validation always fails regardless of API key correctness. The DeepSeek BYOK card exists in the UI, but the backend doesn't know how to handle `"deepseek"` as a model platform.

### Root cause

In `backend/app/model/model_platform.py`, the `PLATFORM_ALIAS_MAPPING` dict is missing an entry for `"deepseek"`. When the UI sends `model_platform: "deepseek"`, the `normalize_model_platform()` function passes it through unchanged (no alias found). CAMEL's `ModelFactory.create()` then receives `"deepseek"` as the platform and throws `ValueError: Unknown model platform`.

The fix is to add `"deepseek": "openai-compatible-model"` — DeepSeek's API is OpenAI-compatible, so it should use the same platform type as `"grok"`, `"ModelArk"`, and `"llama.cpp"` which are already mapped correctly.

### Additional affected providers

While investigating this, I noticed the same class of bug affects several other BYOK providers listed in the UI (`src/lib/llm.ts`). Here's the full audit:

| Provider `id` from UI | Current `PLATFORM_ALIAS_MAPPING` | Resolves to | Status |
|---|---|---|---|
| `deepseek` | *(missing)* | `"deepseek"` (self) | ❌ BROKEN — should be `"openai-compatible-model"` |
| `z.ai` | `"zhipu"` | `"zhipu"` | ❌ BROKEN (#1556) — should be `"openai-compatible-model"` (CAMEL doesn't support "zhipu") |
| `ernie` | `"qianfan"` | `"qianfan"` | ❌ Possibly broken — needs verification that CAMEL supports "qianfan" |
| `minimax` | *(missing)* | `"minimax"` (self) | ❌ Possibly broken — needs verification |
[...]

Issue #1556, opened 2026-04-13
[BUG] 'Unknown model platform: zhipu' error when configuring Z.ai as BYOK custom model


supermomonga:
### What version of eigent are you using?

latest (as of 2026-04-13)

### System information

Reproducible on macOS and Linux (please specify actual environment if different). Happens when setting a Z.ai API key as a BYOK provider in Eigent's model configuration UI.

### Problem description

**Short Description:**
When attempting to configure a custom model for Z.ai (zhipu), validation fails with the error message: 'Unknown model platform: zhipu'.

**Expected behavior:**
The model platform should be recognized, and API key validation should complete successfully if credentials and endpoint are correct.

**Actual behavior:**
Regardless of API key or URL, configuration fails and returns the error: 'Unknown model platform: zhipu'.

**Repro steps:**
1. Go to Settings > Models (BYOK section)
2. Choose Z.ai
3. Enter a valid API key and endpoint
4. Click Save/Verify
5. Observe error message in the UI

**Relevant error log:**
```plaintext
Unknown model platform: zhipu
```

**Note:** Z.ai is mapped to 'zhipu' in the backend, but it appears 'zhipu' is not recognized by the camel-ai model backend. This causes validation to always fail.

### log

(See error above. Additional logs can be attached if required.)

### Suspected root cause
- In `backend/app/model/model_platform.py`, Z.ai is aliased to 'zhipu'. However, camel-ai ModelFactory (or ModelPlatformType) does not support 'zhipu' as a model platform, leading to the error.
- Possibly missing mapping or incomplete support for Zhipu in camel-ai backend.

### Suggested solution
- Either update the aliasing so 'z.ai' maps to an OpenAI-compatible platform (if appropriate),
- OR add explicit support for 'zhipu' in the camel-ai backend if Zhipu models should be supported natively.

#### Related code reference
- See the PLATFORM_ALIAS_MAPPING in backend/app/model/model_platform.py

### Screenshots
Attach screenshots of the settings page and error message if possible.

supermomonga:
@xinmotlanthua Thanks, that really helps!

davidemarchisio:
This is the same class of bug as #1578 — providers mapped to values that CAMEL's `ModelPlatformType` doesn't recognize. In this case `"z.ai"` is aliased to `"zhipu"` which doesn't exist in CAMEL's enum, just like `"deepseek"` passes through unmapped entirely. Both should be mapped to `"openai-compatible-model"` instead.

PR #1675, opened 2026-06-09
fix: add DeepSeek, Z.ai, MiniMax, Moonshot, SambaNova to PLATFORM_ALIAS_MAPPING (fixes #1578, #1556)

carlosjarenom:
## Problem

Users configuring DeepSeek, Z.ai, MiniMax, Moonshot, or SambaNova as custom API keys through the BYOK UI would encounter validation failures. The backend's `PLATFORM_ALIAS_MAPPING` was missing these providers.

Related issues: #1578 (DeepSeek), #1556 (Z.ai)

## Root Cause

The frontend (`src/lib/llm.ts`) lists these providers as BYOK options, but the backend alias mapping didn't include entries for them, so they were passed directly to CAMEL which couldn't recognize them.

Additionally, `z.ai` was incorrectly mapped to `"zhipu"` instead of `"openai-compatible-model"`.

## Fix

Updated `backend/app/model/model_platform.py`:

- `deepseek` → `openai-compatible-model` (new)
- `z.ai` → `openai-compatible-model` (was incorrectly `zhipu`)
- `minimax` → `openai-compatible-model` (new)
- `moonshot` → `openai-compatible-model` (new)
- `samba-nova` → `openai-compatible-model` (new)

All these providers expose OpenAI-compatible API endpoints.

Also updated the test file to cover the new mappings.

## Files Changed

- `backend/app/model/model_platform.py` - Added missing provider aliases
- `backend/tests/app/model/test_model_platform.py` - Updated tests

PR #1781, opened 2026-07-19
fix: map z.ai and deepseek to openai-compatible-model (#1556, #1578)

carlosjarenom:
## Changes

- **backend/app/model/model_platform.py**: 
  - Changed `z.ai` alias from `zhipuai` → `openai-compatible-model`
  - Added `deepseek` alias → `openai-compatible-model`
  - Both providers use OpenAI-compatible APIs

- **backend/tests/app/model/test_model_platform.py**: Updated assertions for new mappings

## Fixes
- #1556: `Unknown model platform: zhipu` when configuring Z.ai as BYOK
- #1578: `Unknown model platform: deepseek` when configuring DeepSeek BYOK

## Root Cause
CAMEL's `ModelPlatformType` enum doesn't have `zhipu` or `deepseek` values. Both Z.ai and DeepSeek expose OpenAI-compatible endpoints, so they should map to `openai-compatible-model` (like `grok`, `nebius`, `ModelArk`, etc. already do).

### DIFF
--- a/backend/app/model/model_platform.py
+++ b/backend/app/model/model_platform.py
@@ -17,7 +17,8 @@ from typing import Annotated, Final
 from pydantic import BeforeValidator
 
 PLATFORM_ALIAS_MAPPING: Final[dict[str, str]] = {
-    "z.ai": "zhipuai",
+    "z.ai": "openai-compatible-model",
+    "deepseek": "openai-compatible-model",
     "ModelArk": "openai-compatible-model",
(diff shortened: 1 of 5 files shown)


carlosjarenom:
Superseded by #1675 (same fix: z.ai/deepseek PLATFORM_ALIAS_MAPPING). Closing duplicate to reduce review noise. Thanks!

PR #1782, opened 2026-07-19
fix(security): add alembic.ini to .gitignore (#1527)

carlosjarenom:
alembic.ini contains database connection strings and should not be committed.

Part of #1527 security hardening.


carlosjarenom:
Superseded by #1677 (security cleanup including alembic.ini in .gitignore). Closing duplicate. Thanks!

PR #1783, opened 2026-07-19
fix(security): require CHAT_SHARE_SECRET_KEY env var (#1527)

carlosjarenom:
## Changes

- **server/app/model/chat/chat_share.py**: 
  - Remove random fallback for `CHAT_SHARE_SECRET_KEY` - app now fails with clear error if not configured
  - Add warning for missing `CHAT_SHARE_SALT` (can remain optional with warning)

## Fixes
- Part of #1527: Hardcoded fallback key in chat_share.py

## Security Impact
Previously, if `CHAT_SHARE_SECRET_KEY` was not set, a random ephemeral key was generated. This meant share tokens would not survive restarts, but more importantly, the fallback was a silent failure. Now the app refuses to start without the key, forcing proper configuration.

Generate a secure key with:
```bash
openssl rand -base64 32
```

### DIFF
--- a/backend/app/model/model_platform.py
+++ b/backend/app/model/model_platform.py
@@ -17,7 +17,8 @@ from typing import Annotated, Final
 from pydantic import BeforeValidator
 
 PLATFORM_ALIAS_MAPPING: Final[dict[str, str]] = {
-    "z.ai": "zhipuai",
+    "z.ai": "openai-compatible-model",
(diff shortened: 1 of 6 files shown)


carlosjarenom:
Superseded by #1677 (security cleanup including CHAT_SHARE_SECRET_KEY env var). Closing duplicate. Thanks!

Issue #1511, opened 2026-03-23
[BUG] Groq API incompatibility with Browser Agent tool schema


SuKops:
### What version of eigent are you using?

Version: Eigent v0.0.88

### System information

When using Groq API via OpenAI Compatible BYOK configuration, all tasks (including simple chat and writing tasks) are routed through the Browser Agent which fails with a 400 error due to strict JSON schema validation on Groq's side.

### Problem description

BadRequestError: Error code: 400 - {'error': {'message': 'invalid JSON 
schema for tool browser_type, tools[2].function.parameters: 
/properties/inputs/anyOf/0/items: `additionalProperties:false` must be 
set on every object', 'type': 'invalid_request_error', 'param': 'tool 
browser_type, tools[2].function.parameters'}}
(assigned to worker: Browser Agent)

### log

Provider: OpenAI Compatible (Groq)
API Host: https://api.groq.com/openai/v1
Model: llama-3.3-70b-versatile
OS: Windows

### Additional context

Steps to Reproduce:

Configure Groq via BYOK → OpenAI Compatible
Set as Default model
Run any task in Eigent (web search, writing, chat)
Error appears — task fails immediately

Expected Behaviour:
Simple tasks like writing and chat should work without browser tools
Browser Agent should only activate when web browsing is explicitly needed

Actual Behaviour:

All tasks routed to Browser Agent regardless of task type
Groq rejects the tool schema with 400 error
Task fails immediately

PR #1787, opened 2026-07-19
fix(anthropic): coerce max_tokens to int to fix validation error (#1519)

carlosjarenom:
## Changes

- **backend/app/agent/agent_model.py**: Added coercion of `max_tokens` to integer for Anthropic models.

## Root Cause

`max_tokens` can arrive as a string from:
- Environment variables (all env vars are strings)
- JSON config from frontend
- Legacy `extra_params` 

Anthropic API requires `max_tokens` to be an integer, causing 400 error:
```
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'max_tokens: Input should be a valid integer'}}
```

## Fix

Added type coercion with fallback to default (128000) on invalid values:

```python
if "max_tokens" in model_config and isinstance(model_config["max_tokens"], str):
    try:
        model_config["max_tokens"] = int(model_config["max_tokens"])
    except (ValueError, TypeError):
        logger.warning(...)
        model_config["max_tokens"] = 128000
```

## Fixes

- #1519: Anthropic API Key cannot be configured in MacOS desktop app (and Linux/Win)


4pmtong:
Thanks for the contribution and for looking into these issues. @carlosjarenom 

I found a few blocking points:

- The Groq change is outside this PR’s description. Also, #1511 uses Groq through the OpenAI-compatible provider, so the `model_platform == "groq"` check will not run. On the current CAMEL version, the `browser_type` schema is already strict. Please remove this change or submit it separately with a failing regression test.
- The Anthropic conversion only runs when creating the runtime agent. The model settings flow calls `/model/validate` first, where a string `max_tokens` can still fail. Please use shared normalization for both runtime and validation.
- Please do not silently replace an invalid value with `128000`, since token limits differ between Anthropic models. Validation should report invalid values clearly, and any runtime fallback should be conservative.
- Please add regression tests for both the validation and runtime paths.

We’ll keep this PR open for now. If you have time to update it, we’ll be happy to review it again.

PR #1882, opened 2026-09-01
fix: preserve Responses instructions across chained turns

4pmtong:
## Summary
- send the trusted agent prompt through Responses API instructions on every request
- remove duplicate system/developer items from Responses input to avoid double billing
- bump the desktop version to 1.0.4

## Validation
- backend/.venv/bin/pytest backend/tests/app/agent/test_agent_model.py -q (20 passed)
- pre-commit hooks passed

## Rollout
Deploy the server-side Prompt Guard Responses-shape support before publishing v1.0.4.

### DIFF
--- a/backend/app/agent/agent_model.py
+++ b/backend/app/agent/agent_model.py
@@ -57,6 +57,57 @@ _NATIVE_STREAM_USAGE_PLATFORMS = {
 }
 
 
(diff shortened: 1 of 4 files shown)


4pmtong:
Superseded by the same commit on the requested fix/responses-prompt-guard-v1.0.4 branch.

PR #1895, opened 2026-09-03
fix: restrict ~/.eigent env files to owner-only (0600)

### LINKED ISSUE #1670: [Linux] Environment files created without secure permissions
## Description

When eigent creates or updates user-specific .env files, it does not set restrictive file permissions on them.

**Location:** backend/app/component/environment.py - set_user_env_path()

## The Problem

On Linux, files containing API keys, tokens, or credentials should be readable only by the owning user (mode 0600). Currently, when a .env file is created or loaded, no chmod is applied.

This means the file could be readable by other users on the system, depending on the default umask.

## Steps to Reproduce

1. Install eigent and configure API keys in .env
2. Check permissions on the generated ~/.eigent/.env file
3. The file may have default permissions (e.g., 0644) instead of 0600

## Expected Behavior

Sensitive environment files should be created with mode 0600 (owner read/write only).

## Actual Behavior

Files are created with default umask permissions, potentially allowing other users on the system to read credentials.

## System

- **OS:** Linux (Arch Linux)
- **Installation:** git clone + npm install + npm run dev

## Suggestion

Apply os.chmod(path, 0o600) after creating or writing any .env file in the backend, to ensure credentials are only accessible by the owning user.

BetterAndBetterII:
## Summary
Credential-bearing `.env` files under `~/.eigent` were left at the process umask (often `0644`) after Electron `writeFileSync` updates. `getEnvPath` only applied `chmod 0600` on the copy-from-default create path, so later writes stayed world-readable.

Harden create and load paths on POSIX so secrets stay owner-readable:
- Backend: restrict permissions when loading `~/.eigent` env files / resetting to the default path
- Electron: `writeEnvFile` writes with mode `0600` and re-chmods existing files; `getEnvPath` tightens existing user env files

Fixes #1670

## Test plan
- [x] Backend pytest: existing `0644` `~/.eigent` env files become `0600` on load/reset; repo-local `.env` left alone; chmod failure does not block load
- [x] Electron vitest: `getEnvPath` / `writeEnvFile` enforce `0600` on create and update

### DIFF
--- a/backend/app/component/environment.py
+++ b/backend/app/component/environment.py
@@ -40,6 +40,29 @@ env_base_dir = os.path.join(os.path.expanduser("~"), ".eigent")
 default_env_path = os.path.join(env_base_dir, ".env")
 
 
+def _restrict_env_file_permissions(path: str | os.PathLike[str]) -> None:
+    """Restrict a user env file to owner read/write (0600) on POSIX.
+
+    Best-effort: permission errors must not prevent loading credentials.
+    Only files under env_base_dir are changed.
+    """
+    if os.name != "posix":
+        return
+    try:
+        resolved = Path(path).expanduser().resolve()
+        resolved.relative_to(Path(env_base_dir).resolve())
+    except (ValueError, OSError):
+        return
+    try:
+        os.chmod(resolved, 0o600)
+    except OSError as exc:
+        logger.warning(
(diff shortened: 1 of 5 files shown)


4pmtong:
@BetterAndBetterII Thanks for contribution!

